PR #11, opened 2015-01-07
Cdap

fscottfoti:
This PR - which I won't be merging yet - contains the CDAP for number of persons in the households == 1  

To the good: I had to dig out the definitions of some variables from the code and ended up finding everything I needed.  Started a data dictionary on the wiki here: https://github.com/synthicity/activitysim/wiki/Data-Dictionary

Not as good: the multiple member households have some tricky logic in how they choose their CDAP.  I got a good explanation from David Ory and will dig into multi-member households next.


e-lo:
At some point it might be good to (since this is an example) write out exactly what is going on in the comments...in travel modeler ( not programmer shorthand ) terms.

I.e. cdap took me a few minutes to figure out what it was. I had to look at the roadmap and then I was like AHA...why didn't you say so! ( or maybe this is bc we just have the less sophisticated DAPs instead of CDAPs)

If it took me 5 min to figure out...my guess that somebody less familiar ( i.e. an intern!) will take longer.

This isnt a high priority for now, but let's keep it mind when we use shorthand and acronyms.

fscottfoti:
Just a quick reminder, for anyone following this repo - this is what I will be working on this week.  I believe I have the CDAP (coordinate daily activity pattern) running for 1 person households and will now be proceeding to 2 person households.  If you guys would prefer that I be working on something else (e.g. if you don't need the C of CDAP), feel free to let me know.

Elizabeth, your comment about adding comments is important.  I suggest we do a commenting and testing iteration after getting enough of a prototype going that we all think that the current approach is good.  I don't to to do much polish on code that's still subject to significant revisions...

jiffyclub:
Closing this since the PR isn't targeting the master branch.

PR #17, opened 2015-02-20
Workplace location

fscottfoti:
This is the code required to get the workplace location model to work.  It is now up-to-date with the external OMX code and is working and ready to merge.


example/models.py
```diff
@@ -1,6 +1,10 @@
 import urbansim.sim.simulation as sim
 import os
 from activitysim import activitysim as asim
+from activitysim import omx
```
jiffyclub: This is still pulling `omx` from `activitysim`.

example/models.py
```diff
@@ -8,12 +12,34 @@ def auto_alts():
...
 
+@sim.table()
+def zones():
+    # I grant this is a weird idiom but it helps to name the index
+    return pd.DataFrame({"TAZ": np.arange(1454)+1}).set_index("TAZ")
+
+
+@sim.injectable()
+def nonmotskm_omx():
+    return omx.open_omxfile('data/nonmotskm.omx')
```
jiffyclub: This will be `omx.openFile` once you're using [openmatrix](https://pypi.python.org/pypi/OpenMatrix).

PR #29, opened 2015-03-04
File reorg

fscottfoti:
Major file reorganization.  This PR subsumes all the previous PRs.  Also added tests to this - travis is likely to fail the first couple of times through - we shall see.


fscottfoti:
@jiffyclub Any idea why travis isn't running on this PR?

jiffyclub:
Looks like it's getting started now, dunno why it took a while.

fscottfoti:
So this is failing because there's a dependency on those `dcm` changes you made to UrbanSim.  Can we update this repo to depend on master of urbansim or do you think we should do another urbansim release?

jiffyclub:
I don't want to release urbansim just yet because I need to figure out why the sanfran_urbansim test is failing. If you want to install urbansim from master on Travis take it out of the conda installation and add this line at the end of the installation section:

```
- pip install https://github.com/synthicity/urbansim/archive/master.zip
```

fscottfoti:
Sounds good - that should work.

PR #34, opened 2015-03-18
Mode choice

fscottfoti:
Opening a pull request for the mode choice model

Not all that much new stuff has been added here.  The most complicated part was adding a 3rd dimension to skims so that time periods can change on a row-by-row basis.

It should look fairly different from the previous implementation (the CTRAMP one), but this should look very similar to other model specs in activitysim.

Two other moderate changes were added too including 1) being able to inject constants into the variable evaluations and 2) evaluating coefficients in terms of other coefficients.  All of this is performed pretty easily with Python `eval` although it required a pretty diligent reorganization of the spec files from the CTRAMP way.  For instance, coefficients were all set to `1` in the coefficient column and then the actual betas were put in the expression column - I guess this was because coefficients could not be expressions before.


fscottfoti:
I should add that I don't consider this PR complete at this time and won't be merged without some further edits.

activitysim/skim.py
```diff
@@ -49,3 +50,229 @@ def get(self, orig, dest):
...
+        """
+        self.skims = {}
+        self.left_key = "TAZ"
+        self.right_key = "TAZ_r"
+        self.df = None
+
+    def set_keys(self, left_key, right_key):
+        """
+        Set the left and right keys.
+        Parameters
```
jiffyclub: Need a blank line before this heading for Sphinx to parse it correctly.

activitysim/skim.py
```diff
@@ -49,3 +50,229 @@ def get(self, orig, dest):
             dest = dest + self.offset
 
         return self.data[orig, dest]
+
+
+class Skims(object):
+
+    def __init__(self):
+        """
```
jiffyclub: It's kind of inconsistent, but the class docstring doesn't go in the `__init__` method, it goes right after the `class Foo(object):` line. That includes the description of the parameters to `__init__`.

activitysim/defaults/models/mode.py
```diff
@@ -0,0 +1,124 @@
+import os
+import copy
+import yaml
+import pandas as pd
+import urbansim.sim.simulation as sim
+from activitysim import activitysim as asim
+from activitysim import skim as askim
+
+"""
+Mode choice is run for all tours to determine the transportation mode that
```
e-lo: Just a note that mode choice happens for tours and trips, so eventually you will either be generic in this file to accept either or create something very similar (in which case you may want to rename this file).  This isn't urgent...just something to think about.

activitysim/defaults/models/workplace_location.py
```diff
@@ -56,21 +56,23 @@ def workplace_size_terms(land_use, destination_choice_size_terms):
...
     alternatives = zones.to_frame().join(workplace_size_terms.to_frame())
 
-    skims = {
-        "distance": distance_skim
-    }
+    # set the keys for this lookup - in this case there is a TAZ in the choosers
+    # and a TAZ in the alternatives which get merged during interaction
+    skims.set_keys("TAZ", "TAZ_r")
+    # the skims will be available under the name "skims" for any @ expressions
+    locals_d = {"skims": skims}
```
e-lo: I'm not quite sure i understand the benefit of this extra layer of dict?  (and apologies for being a n00b)

activitysim/defaults/tables/households.py
```diff
@@ -15,6 +15,14 @@ def households(store, settings):
     return store["households"]
 
 
+# this is a common merge so might as well define it once here and use it
+@sim.table()
+def households_merged(households, land_use, accessibility):
```
e-lo: "merged" is a little generic.

activitysim/defaults/tables/skims.py
```diff
@@ -39,3 +39,17 @@ def sovmd_skim(nonmotskm_matrix):
 def sovpm_skim(nonmotskm_matrix):
     # FIXME use the right omx file
     return skim.Skim(nonmotskm_matrix, offset=-1)
+
+
+@sim.injectable(cache=True)
+def skims():
+    skims = skim.Skims()
+    # FIXME - this is reusing the same skim as all the different kinds of skims
+    for typ in ["SOV_TIME", "SOVTOLL_TIME", "HOV2_TIME",
```
e-lo: This is for later, but it would be good to define these at the top of a file or in a spec rather than inline in the code.  It would just be good to be super obvious about stuff like this that could change.

e-lo: Also - modes and variables w/in modes DO change and can be different for TOUR modes and TRIP modes.

activitysim/defaults/models/workplace_location.py
fscottfoti: Mostly because it's a choice of hardcoding inside the simulation that skims is available under the name "skims"... or the alternative, which is to allow the user to pass in anything they want with any names they prefer (which can now be referred to in the expressions in the CSV file).

This is used more liberally at the link below (in mode choice), where there is a separate skims object available for the trip away from home as the trip back home, and all of the constants which are stored in YAML are added to the scope in as well (and there are plenty of constants in the mode choice).  

But really it's an attempt to keep simple_simulate with as little custom logic as possible.  Although skims are fairly common, they're certainly not used by all models, and it's unfortunate that they have to be passed in at all.  The only thing they're used for is to add the merged dataframe (of choosers and alternatives) back to the skims object.  I think this could be fixed by having a "model_prep" function but this kind of abstraction might be difficult to follow for domain experts.  I find programming to be a constant push-and-pull from abstraction to domain-specificity and the implementation choices are always a question of taste.  Really I could be talked into going either direction (towards abstraction or towards domain specific).

https://github.com/synthicity/activitysim/pull/34/files#diff-800bb6be26eb1bf48d40b8ccf7e48facR106

activitysim/defaults/tables/skims.py
fscottfoti: In this specific case, I'm really just reading in the distance skim 15 times so that I have skims to play with ;)  But in a more general situation, I think it's interesting to know what is "core" and what is "configuration."  I'm not certain there's a bright red line.  I mean, in many cases I think it's easier to  "configure" stuff directly in Python rather than having to go to a second location to find out what something is - most functions are now at most 15 lines long and having all the code in one place is pretty nice.  On the other hand if something is used more than once, it should probably be defined once and shared, whether it's defined in Python or a text file.  

So far I've set a principle for myself that models should be very generic, should be shared among regions, should only be written by a certain class of users/programmers, etc.  But tables (data sources) and columns mostly just use Pandas and I think should be fair game for a much larger class of users.  As long as a variable is defined only once (which the simulation framework is designed to enable exactly that), I think it's fair game to "configure" a variable in Python.

But this principle is still sort of a work in progress and is perhaps not correct.  Or perhaps it's just not clear that's what is intended.  I think it would require moving the stuff in the "models" directory into the "activitysim" directory and keeping tables/column under defaults.  It is interesting to me to know how much (if any?) of the the variable definitions are shared among regions.  I'm guessing not much.  But I'm sure models can run the same way and can be shared between regions with no edits in the code.

Again, this was true of CTRAMP in a major way as well.  There are constant references to methods in the code for things that can't be computed in the CSV files.  In our version it's just that those Python definitions are easier to write, are shorter, and don't require compilation, so are a lot more accessible to the modelers.  I'm trying to keep all of these things under the "tables" directory, but a few methods should maybe be moved up the stack or down for sure.

fscottfoti:
I see we now have regression tests on a few of the choices.   This is an interesting conundrum, as since these are stochastic models I can't imagine any change I make not breaking these tests.  Should we at least compare the probabilities for changes rather than the choices?  In general though, if we want to test results we should probably use some simplified specs that don't change as often as the examples will (which will be frequent for some time yet).  I suppose I can take on the task of creating some simple specs that won't change for testing.

jiffyclub:
I added in some random number seeding so that draws will turn out the same across runs, but yeah things like changing the spec or the number of choosers will cause the results to change. Simplified test specs are a good idea because they are a lot easier to troubleshoot than production specs with hundreds of variables/alternatives.

jiffyclub:
Deleting those tests is also an option. I put them in because I wanted to check things while I was changing the MNL guts, but if they'll be a nuisance we can take them out.

activitysim/defaults/tables/skims.py
e-lo: I think the main thing is that we don't want to force TOUR and TRIP skims to be the same...even though they might be the same variables, they aren't the same files.

fscottfoti:
@jiffyclub It appears that the regression tests are still not returning the same results every time (they're still random).  I'm setting the seed here - https://github.com/synthicity/activitysim/blob/mode-choice/activitysim/defaults/test/test_defaults.py#L65.  Do you see anything wrong with that code?

jiffyclub:
It looks like in https://github.com/synthicity/activitysim/commit/c288cdcbddcea68a92fb039e4b4fe2e9013ddcc9#diff-6527e38e0f07aa1692422aeca4e28c8b I accidentally removed the `set_random_seed` injectable from `auto_ownership_simulate`. Adding that back in might help.

PR #36, opened 2015-03-20
Adding MNL Utilities to ActivitySim

jiffyclub:
This adds some specialized MNL utilities to ActivitySim to supplement those in UrbanSim. These make it a bit easier to work with the alternative-specific utilities in ActivitySim models and are targeted only at making choices, since at the moment we always have pre-defined coefficients for ActivitySim choice models.


jiffyclub:
Update: I think this is now ready for review and merging.

fscottfoti:
Excellent - I will merge into mode choice and see if I can update to match the new way of doing things.  Looks pretty darn parsimonious to me, just the way I like it.

PR #37, opened 2015-03-20
CDAP Utilities

jiffyclub:
These are utilities for processing CDAP specs. For example, accumulating individual utilities across alternatives according to the household occupants, and then creating alternatives and utilities for each household from the utilities of its occupants.

Remaining work here:

- [x] finish `apply_all_people` function for adding final corrections to household utilities
- [x] create function for making the household choices
- [x] create function for mapping household choices to individuals
- [x] wrap all the CDAP steps into one containing function

There is some overlap between this PR and #36. #36 should be merged first.


jiffyclub:
I think this is ready to go.

fscottfoti:
Awesome - I will try it out tomorrow

jiffyclub:
One thing to take a look at are the test spec files, because not everything is expressed in the same way as in the MTC model: https://github.com/synthicity/activitysim/tree/cdap/activitysim/cdap/tests/data In particular I think the [final rules](https://github.com/synthicity/activitysim/blob/cdap/activitysim/cdap/tests/data/cdap_final_rules.csv) and [all_people](https://github.com/synthicity/activitysim/blob/cdap/activitysim/cdap/tests/data/cdap_all_people.csv) specs might be a little different.

The final_rules are used to set household utilities to a specific value. In the example spec any household has a person that matches `(age == 38) & (income == 1900)` will have all alternatives where that person has the `Mandatory` state set to 0.

fscottfoti:
Just to confirm - for the final rules, the Expression runs for the persons table and sets the specified alternative to a certain probability?  It is probability, right?

jiffyclub:
Good question! No, right now it's utility. We can fix that.

jiffyclub:
Or if it makes more sense to leave it as utilities so it matches the other specs you can put in -999 and we'll fix the docs so it makes sense.

PR #39, opened 2015-03-31
size terms are now fairly complete

fscottfoti:
Size terms are added as a table in the tables directory with one helper function which has a unit test under defaults.  Size terms are now used in the workplace location and destination choice models.  The destination choice models no longer use sampling, so are fairly close to final models.


fscottfoti:
@jiffyclub This is passing on my laptop but is getting "killed" consistently by Travis.  Do you think it's something legitimate or just wait a bit and try again with Travis?

jiffyclub:
The same thing is happening on UrbanSim, guess it's time to troubleshoot this.

jiffyclub:
Based on this post it might be that we're using up all the memory: http://stackoverflow.com/questions/27564156/home-travis-build-sh-line-41-pid-killed-exit-code-137 Which I don't know there's anything we can do about that except try to use less memory for our tests.

fscottfoti:
So 4gb is the limit?  Or is 3gb?

jiffyclub:
I wish I knew. I asked in the Travis-CI IRC but no one responded. Trying Twitter.

jiffyclub:
We may have to stop running simulations as part of the Travis runs unless we can think of a way to make the simulations use less memory.

fscottfoti:
Running with fewer households seems like an option...

fscottfoti:
I can try it and see if it fixes things

jiffyclub:
It was always kind of mysterious as to why the UrbanSim runs stopped working because I don't think we changed anything in the sanfran_urbansim repo. Travis might have lowered the limit without warning.

fscottfoti:
Seems to have fixed the tests by reducing memory use.  Now running with 50 households instead of 1000.

jiffyclub:
That's great! Maybe we can do something similar in sanfran_urbansim?

fscottfoti:
Should be able to.  But 50 records is hardly a test of the simulation - still, better than nothing.

PR #41, opened 2015-04-01
new other_than function

### LINKED ISSUE #22: "other than modeled person" fomulation
@jiffyclub I think this one is for you.  I'm working my way through hundreds of variable specifications, many of which are straightforward.  I now see about 20 that all have the same formulation that has me a bit stumped.  Something along the lines of `a [retiree] in the household that is NOT the current person`.  What do you think is the best way (pandas and/or simulation framework) to formulate a variable like this from the perspective of the chooser?

jiffyclub:
Allows you to test for the presence of something- or someone-else
within a group that has some property. For example, you can figure
out whether there is someone else within a household that has
a given person-type.

Fixes #22.

Using this, the current [`presence_of`](https://github.com/synthicity/activitysim/blob/254da40fe1acd175819533ffcb528a6e5d72053b/activitysim/defaults/tables/households.py#L166) function could be converted to:

```python
def presence_of(ptype, persons, at_home=False):
    if at_home:
        # if at_home, they need to be of given type AND at home
        bools = (persons.ptype_cat == ptype) & (persons.cdap_activity == "H")]
    else:
        bools = persons.ptype_cat == ptype

    return other_than(persons.household_id, bools)
```

Of course, the above `presence_of` would have to go on the `persons` table, where the linked one is on the `households` table.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ install:
 - >
   conda create -q -c synthicity -n test-environment
   python=$TRAVIS_PYTHON_VERSION
-  ipython-notebook jinja2 matplotlib numpy pandas pandana patsy pip pytables
-  pytest pyyaml scipy statsmodels toolz
+  cytoolz ipython-notebook jinja2 matplotlib numpy pandas pandana patsy pip
+  pytables pytest pyyaml scipy statsmodels toolz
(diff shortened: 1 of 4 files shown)


fscottfoti:
Nice - perfect.

PR #43, opened 2015-04-03
new mnl interaction function

jiffyclub:
Replaces the one we were using from UrbanSim. This one is specialized only for simulation and would not support creating an interaction dataset for estimation.


jiffyclub:
@fscottfoti I think we could also merge this one.

fscottfoti:
Sure let's do it.

PR #44, opened 2015-04-03
allow nans in skim orig/dest indexes

jiffyclub:
A nan is returned as the impedence value any time either the origin or destination is nan.


fscottfoti:
Cool - shall we merge this?

PR #55, opened 2015-05-15
Trip mode choice

fscottfoti:
This gets the trip mode choice model as fully implemented as the tour mode choice model is.  There were definitely some additional complications here, especially that the inbound and outbound trips are separated in trip mode choice more than in the tour mode choice.   Although many variables are still unimplemented, the basic structure should be complete, and the spec is filled in for all 350 or so rows in the Excel file.


fscottfoti:
@jiffyclub these tests pass on my laptop - any idea why these are failing on Travis (including some failed tests in cdap)?

jiffyclub:
I'm seeing the same results locally. I wonder if the latest Pandas release changed something.

jiffyclub:
For the cdap tests the failures are coming from this oddness: https://github.com/pydata/pandas/issues/10106#issuecomment-102547092. For the other tests it looks like suddenly the indexes of the things being compared suddenly have names and the reference results don't. I don't know they should have been there all along or what, but they are there now.

jiffyclub:
I pushed a fix for the cdap tests.

jiffyclub:
Ah, maybe it's not the index that needs to be named, but the Series itself.

jiffyclub:
No, it is the index that needs to be named. Like so:

```
(Pdb++) pd.Series([2, 1, 1], index=pd.Index([2306822, 652072, 651907], name='HHID'))
HHID
2306822    2
652072     1
651907     1
dtype: int64
```

fscottfoti:
Ahh yes, I was naming the series.  Seems to be working...

bstabler:
@fscottfoti - what's the status of this?  It seems we should resolve the conflicts and merge this in since it adds the draft trip mode choice model.

fscottfoti:
That's a very good question.  I can think of no reason why it can't be merged - the diffs look pretty standard and I see nothing in the docs I wrote that point out weaknesses.

PR #58, opened 2015-06-08
Fixed bug in female column calculation.

bhargavasana:


fscottfoti:
@jiffyclub we can merge this even though the tests are failing because this is the same issue on master right now?

Issue #116, opened 2016-09-09
correct CDAP


bstabler:
CDAP is supposed to only consider the interactions between up to 5 HH members and then apply some additional utility terms after that.  We will review the MTC TM1 code and UECs and correct this.  CDAP also currently loops by HHs but this is inefficient.

We will likely re-implement it as a series of batch vectorized calculations:
  - Select the 5 persons for consideration in HHs with 6+ people; select workers and youngest kids
  - Calculate person level utilities
  - Calculate HH level utilities
  - Calculate Person pairwise utilities for HH size 1
  - Calculate Person pairwise utilities for HH size 2
  - Calculate Person pairwise utilities for HH size 3
  - Calculate Person pairwise utilities for HH size 4
  - Calculate Person pairwise utilities for HH size 5
  - Sum up utilities at the HH level

The key is to organize the problem into a series of batch table operations

toliwaga:
According to the documentation for applyModelForExtraHhMembers in mtctm2.abm.ctramp HouseholdCoordinatedDailyActivityPatternModel.java choices for extra household members should be made according to a table of fixed proportionate probabilities based on person type:

```
* Applies a simple choice from fixed proportions by person type for members of
* households with more than 5 people who are not included in the CDAP model. The
* choices of the additional household members are independent of each other.
```

Why does it do this rather than choosing activities for the extra members by applying the individual utilities, ignoring the interaction utilities?

It is easy for me to implement this either way...

bstabler:
Here are the CDAP 6+ persons proportions from MTC TM1 
 
   public final double[][] CDAP_6_PLUS_PROPORTIONS = {
            { 0.0, 0.0, 0.0 },
            { 0.79647, 0.09368, 0.10985 },
            { 0.61678, 0.25757, 0.12565 },
            { 0.69229, 0.15641, 0.15130 },
            { 0.00000, 0.67169, 0.32831 },
            { 0.00000, 0.54295, 0.45705 },
            { 0.77609, 0.06004, 0.16387 },
            { 0.68514, 0.09144, 0.22342 },
            { 0.14056, 0.06512, 0.79432 }
        };

bstabler:
We implemented this in a slightly different way - https://github.com/UDST/activitysim/wiki/Project-Meeting-2016.09.23

wusun2:
[...]

PR #118, opened 2016-10-06
rewritten cdap processor

toliwaga:
Completely rewritten cdap processor following strategy outlined in #116 reimplemented cdap as a series of batch vectorized calculations


fscottfoti:
Awesome!  How much faster is it??

toliwaga:
@fscottfoti - benchmarking it now. I think it is faster. Also, since the previous version didn't implement the max hhsize 5 cutoff, it ran out of memory trying to handle households with lots of members. For a household with 14 members, there are 3^14 alternatives, which makes for 4,782,969 columns or alternatives, which is a stretch for pandas!

PR #126, opened 2016-11-05
add cdap to tutorial, clean-up settings to run example

bstabler:
#120


bstabler:
@toliwaga - looks the same bug - https://github.com/UDST/activitysim/commit/bda0328ce8e099af74b64e9c69e39c6f17f5a152 - is getting in the way.  I need to go for today so we'll need to fix this later.

PR #151, opened 2017-02-04
New script

bstabler:
Convert SANDAG network los files to ActivitySim NetworkLOS format


bstabler:
In relation to [multiple zone system support](https://github.com/UDST/activitysim/wiki/Multiple-Zone-Systems-Design)

wusun2:
Is it possible to store MAZtoMAZ data in OMX instead of HDF5?   Maybe even MAZtoTAP data, which are not square matrices?

bstabler:
OMX must be square matrices so the MAZ to MAZ data could be converted to OMX.  The MAZ to TAP data isn't square.  We can convert the MAZ to MAZ data to big sparse matrices, but we just need to be careful about memory use.  I like the idea and I tried it for the CMAP ABM since we only had one MAZ to MAZ measure - distance.  How many measures, or skims, would we need for the SANDAG ABM eventually?

toliwaga:
We could perhaps use sparse matrices? Though I haven't researched this
yet...)

https://docs.scipy.org/doc/scipy-0.18.1/reference/sparse.html

bettinardi:
1-2 tomorrow is available for me

Alex Bettinardi, P.E.
503.986.4104
http://www.oregon.gov/ODOT/TD/TP/Pages/index.aspx

wusun2:
Currently we use 5 MAZtoMAZ impedances: actual walk time, perceived walk time, elevation gain, bike logsum, and bike time.   Although it is hard to predict 'eventually' how many more MAZtoMAZ will be added, it won't be significantly more than the current 5.  

Both walk and bike modes have a cut-off distance threshold to disallow unreasonable long walk and bike trips.  Sparse matrices might be a good fit for walk and bike impedances.

bstabler:
Internally we can use sparse matrices if we want.  How we represent the data structures within the system is up to us.  Externally, ActivitySim currently only supports OMX and pandas table in HDF5.

PR #180, opened 2017-04-27
Asim core

toliwaga:


bstabler:
we're not quite ready to accept this pull request since we need to update the docs and do a little more housekeeping, but I want to keep it open so we can discuss the improvements

e-lo:
Re merging or not the pull request, it seems like it might be time to implement git-flow whereby there is a develop branch so that this can sit there while documentation is updated.  Leaving pull requests open for a long time is less desirable from my perspective.

Issue #235, opened 2018-04-23
add flow chart of the example model sub-models 


bstabler:
@JilanChen suggested

JilanChen:
Thank you

Issue #234, opened 2018-04-23
post simple install / run example script


bstabler:
The steps to install and run the example are on the [Getting Started](https://activitysim.github.io/activitysim/gettingstarted.html) page and the [Example](https://activitysim.github.io/activitysim/abmexample.html#running-the-example-model) page.  We should include in the repo a simple install and run script so the user doesn't have to refer back to the narrative every time.

bstabler:
we decided to streamline the [Getting Started](https://activitysim.github.io/activitysim/gettingstarted.html) text instead

Issue #231, opened 2018-03-30
Error message misreports config dir as output and data dirs


lmz:
https://github.com/ActivitySim/activitysim/blob/b9c51b94e3dae678eb493957e0c3bb307216fde8/activitysim/core/config.py#L48

https://github.com/ActivitySim/activitysim/blob/b9c51b94e3dae678eb493957e0c3bb307216fde8/activitysim/core/config.py#L52

toliwaga:
Oops! Thanks Lisa. I fixed in the joint-tours I am working on, which I expect will be ready for release in a few days.

Issue #223, opened 2018-02-16
small hh sample size causes problems


toliwaga:
For small hh sample sizes (< 100) chooser sets can be empty causing problems for some chunkers and models. Models should be able to handle empty chooser populations for (e.g.) university students.

bstabler:
I think we fixed this?

toliwaga:
Yes, but in a branch that hasn't been pulled to master yet.

PR #240, opened 2018-06-08
Trip scheduling

bstabler:


bstabler:
fix #235, #234, #231, #223, #168, some of #138, some of #137, #135

Issue #273, opened 2019-12-05
Hard Coded Specification - Mandatory Tour Scheduling


danielsclint:
The specifications for mandatory tour scheduling are hard-coded forcing the school and university specification to use the same specification file. In the Atlanta implementation, school and university use different specifications.

https://github.com/ActivitySim/activitysim/blob/7b57c94e12d27a23f86c8988117e885ae754c44e/activitysim/abm/models/mandatory_scheduling.py#L77-L83

It would nice if the model treated the spec more flexibly similar to the way the the location models define the spec. 

https://github.com/ActivitySim/activitysim/blob/7b57c94e12d27a23f86c8988117e885ae754c44e/activitysim/abm/models/location_choice.py#L251

I could imagine a situation where the mandatory specs are defined in the mandatory_tour_scheduling.yaml as follow:

```
SPEC:
  work: tour_scheduling_work.csv
  school: tour_scheduling_school.csv
  univ: tour_scheduling_university.csv
```

PR #275, opened 2019-12-05
Provide more flexibility for defining mandatory schedule specifications.

danielsclint:
#273. This change require defining tour scheduling specs specifically in the mandatory_tour_scheduling.yaml. The code still hard-codes the 'univ' value, and additional flexibility should be added in the future. This is step one of probably a couple of steps to unwind the hard-coded values in this section of the code. 

```
SPEC:
  work: tour_scheduling_work.csv
  school: tour_scheduling_school.csv
  univ: tour_scheduling_university.csv
```
# Review Criteria Responses
_The example scripts have been updated so the examples and tests complete. It's unclear how the documentation should be updated for this change. Few of the internal workings of model settings are well-documented, and it seems like a large discussion (or work effort) is needed to standardize this across the models._

_Yes, this change improves the flexibility of the ActivitySim framework allowing it be implemented in more locations._

_This change will have no material change on the model run time. This code change replaces hard-coded values with a set of user-defined parameters._

_This change does not impact the data pipeline._

_No regression testing has been done. The code change removes a hard-coded value in the Python code with a set of user-defined variables. If the user specifies the same values as the previously hard-coded values, they **should** get the same results. The unit tests seem to confirm this assertion._

_The test configuration files were modified to use the newest features._

_The documentation is still being built out, so it is difficult to understand where and how to add new documentation into the existing framework. Some documentation should be consistent across all models (as is the case with this change), so a standard would need to be set by the larger team or asserted by me._


danielsclint:
@bstabler, I updated the original PR comment with the some additional documentation and answered the "Review Criteria" questions. This PR works, but as I note in the comment, it probably should be followed-up with a few more code enhancements to make the assertion of the 'univ' trips more flexible.

bstabler:
This looks good and we accept the PR.  I changed the target to the develop branch so we can merge features here before merging (releasing) to master.

Issue #285, opened 2019-12-20
Large Matrices Result in Integer Overflow


danielsclint:
I thought this problem was eradicated with PR #265. Unfortunately, with really large numbers, the issue still occurs.

https://github.com/ActivitySim/activitysim/blob/7b57c94e12d27a23f86c8988117e885ae754c44e/activitysim/abm/tables/skims.py#L186-L194

The issue is in the [Numpy library np.prod() method](https://docs.scipy.org/doc/numpy/reference/generated/numpy.prod.html). According to the documentation:

> Arithmetic is modular when using integer types, and no error is raised on overflow.

```python
>>> x = np.array([536870910, 536870910, 536870910, 536870910])
>>> np.prod(x)
16 # may vary
```

danielsclint:
Fixed with PR #286.

PR #286, opened 2019-12-20
Fix Integer Overflow

danielsclint:
This address #285 to fix overflows when calculating the memory footprint for the skims. This replaces the Numpy.prod method with a straight Python integer multiplication to let Python handle the growth of the integer value under the hood.

This conversation on [Stack Overflow](https://stackoverflow.com/questions/39089618/why-is-numpy-prod-incorrectly-returning-negative-results-or-0-for-my-long-li) provides a good synopsis of the problem and the solution.

Rather than replace Numpy prod in line with the multiplication, I created a short method to do multiplication that might result in big numbers. This understandably may not be the most elegant solution, but it does allow me to write a test around the problem.

```python
def multiply_large_numbers(list_of_numbers):
    return reduce(mul, list_of_numbers)
```

# Review Criteria Responses
_This code doesn't change the examples or documentation. It is a background fix to eradicate a bug. A test has been written to prove that it works._

_Yes. It allows for ActivitySim to support much larger implementations._

_This change has no material impact on runtimes._

_This change does not impact the data pipeline._

_No regression testing has been done. The code change fix a bug that results in incorrect arithmetic in certain situations._

_Yes, a test was written along with scaffolding to test other methods in the skims.py file._

_No_


bstabler:
@danielsclint - please rebase this against the develop branch so it incorporates #275

danielsclint:
I updated the target branch to develop.

PR #323, opened 2020-06-29
Fix min inbound departure times

mxndrwgrdnr:
Updates to two modules to fix the issue in which the first inbound trip of a given tour is scheduled to depart prior to the last outbound trip of the tour:

- models/trip_scheduling.py -- new method to update `earliest` column in trips table after outbound trips are scheduled
- models/util/trip.py -- updates def of `failed_trip_cohorts`. cohort must include _all_ trips associated with a tour that has a failed trip because of the new interdependence between inbound and outbound departure times.

### DIFF
--- a/activitysim/abm/models/trip_scheduling.py
+++ b/activitysim/abm/models/trip_scheduling.py
@@ -88,6 +88,40 @@ def set_tour_hour(trips, tours):
         reindex(subtours[subtours.tour_num == subtours.tour_count]['end'], trips[inbound].tour_id)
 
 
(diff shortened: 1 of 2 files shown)


bstabler:
Hi @mxndrwgrdnr, thanks for the PR.  The master branch is our current release branch and is only pushed to by Travis once develop passes the tests.  In addition, since this PR has some revisions already in the RSGInc fork develop branch, I think it is best to submit this PR to the RSGInc fork develop branch.  I'm going to close this PR and then we can follow-up on the new PR.  Thanks.

PR #357, opened 2020-11-09
Updated joint_tour_destination configs

albabnoor:
Updated joint_tour_destination.py script to point to joint_tour_destination.yaml. Added joint_tour_destination model configs files which are now separated from the non-mandatory destination choice model configs.


bstabler:
Thanks @albabnoor.  Instead of creating a separate (and somewhat duplicate) set of inputs for this model, we added a ```include_settings``` option that clarifies that this submodel shares a configuration file with another submodel.  The revised [joint_tour_destination.yaml](https://github.com/ActivitySim/activitysim/blob/develop/activitysim/examples/example_mtc/configs/joint_tour_destination.yaml) now simply includes the following - ```include_settings: non_mandatory_tour_destination.yaml```.

PR #365, opened 2021-01-04
ARC Updates

danielsclint:
Initial PR for ARC Updates

### DIFF
--- a/activitysim/abm/models/__init__.py
+++ b/activitysim/abm/models/__init__.py
@@ -29,6 +29,7 @@ from . import trip_destination
 from . import trip_mode_choice
 from . import trip_purpose
 from . import trip_purpose_and_destination
+from . import trip_scheduling
 from . import trip_departure_choice
 from . import trip_scheduling_choice
 from . import trip_matrices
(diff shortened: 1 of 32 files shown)


bstabler:
Thanks @danielsclint.  I'm going to accept and then add an additional example to exercise the ARC extensions to TM1.  This will take the form of a new entry here - https://github.com/ActivitySim/activitysim/blob/develop/activitysim/examples/example_manifest.yaml.

bstabler:
@danielsclint - the `example_mtc_arc_extensions ` on `develop` fails in one of the new ARC submodels.  Please fix so we can release the next version.  Here's how to create and run this new example.  Thanks. 

```
activitysim create -e example_mtc_arc_extensions -d test_example_mtc_arc_extensions
activitysim run -c configs_arc -c configs -o output -d data
```

danielsclint:
You defined trip_scheduling and the two ARC trip_scheduling models to run in the same run.

image

You should only define one or the other.

image

bstabler:
Thanks.  It is now fixed.  This is a good example of why we need to include runnable examples with contributions.

PR #368, opened 2021-01-14
Added work from home and telecommute frequency models.

albabnoor:
also updated cdap specs to reflect telecommute choices


bstabler:
accepting this work in progress

bstabler:
see https://github.com/ActivitySim/activitysim/wiki/Telecommute

PR #372, opened 2021-01-21
Write Number of Participants to Trip List Output

danielsclint:
As a convenience method, the ARC Cube scripts need a way to expand the output trip lists based on the number of participants. This pull request adds an additional field "number_of_participants" to the final_trips.csv. It is a mirror of the similar value on the final_tours.csv. 

This PR also removes a vestigal time_period_label method that had been replaced the new Network LOS class.

# Review Criteria Responses
_The examples are updated with the appropriate configuration changes. No new documentation is necessary. No new tests were created, but the existing tests all pass._

_Yes, it provides more flexibility to support trip table expansion without necessitating table joins. This extends the spirit of the ActivitySim project to provide maximum implementation flexibility while maintaining a common workflow and model specification._

_This change has no material impact on runtimes._

_This change requires a change to the data pipeline to deal with redundant field names in downstream models. The examples have been updated._

_No regression testing has been done. The code change removes a hard-coded value in the Python code with a set of user-defined variables. If the user specifies the same values as the previously hard-coded values, they **should** get the same results. The unit tests seem to confirm this assertion._

_The test configuration files were modified to use the newest features._

_No. This is pretty straightforward._

### DIFF
--- a/activitysim/abm/models/stop_frequency.py
+++ b/activitysim/abm/models/stop_frequency.py
@@ -82,23 +82,24 @@ def process_trips(tours, stop_frequency_alts):
 
     trips['person_id'] = reindex(tours.person_id, trips.tour_id)
     trips['household_id'] = reindex(tours.household_id, trips.tour_id)
+    trips['number_of_participants'] = reindex(tours.number_of_participants, trips.tour_id)
 
     trips['primary_purpose'] = reindex(tours.primary_purpose, trips.tour_id)
 
     # reorder columns and drop 'direction'
     trips = trips[['person_id', 'household_id', 'tour_id', 'primary_purpose',
-                   'trip_num', 'outbound', 'trip_count']]
+                   'trip_num', 'outbound', 'trip_count', 'number_of_participants']]
 
     """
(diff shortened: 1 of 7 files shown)


guyrousseau:
This PR also removes a vestigial ...

bstabler:
thanks @danielsclint.  The pipeline table `joint_tour_participants` contains the number of participants and can be output via the settings file `output_tables` list and then table joined later if needed.  I think this is a more elegant solution than adding `REDUNDANT_TOURS_MERGED_CHOOSER_COLUMNS` directives to submodels to deal with additional data we added upstream but don't need until we're using the outputs in an external process.  Thoughts?

bstabler:
- @JilanChen suggested adding the the field you suggest as a good way to help document what's in the trips table.
  - @joecastiglione suggested we make it configurable to write different types of trips tables - raw trips, expanded person level trips, etc.

bstabler:
Another idea discussed is to improve the output trip writer to support annotations as a way to support more flexible/custom output writing

PR #389, opened 2021-03-05
Document input files for MTC TM1 example

mattwigway:
This adds documentation for all the fields and codes in the MTC TM1 input files, based on reading constants.yml and [the Bay Area Metro data dictionaries](https://github.com/BayAreaMetro/modeling-website/wiki/DataDictionary). If any of these codes have changed for the ActivitySim implementation of TM1, these could be incorrect - but they seem to match up.

I based this on the develop2 branch since that seems to be where development is occurring, but I can change the base if that's not correct.


bstabler:
@mattwigway - this is great, thanks for this.  I'll merge it soon.  Let me know when you are done.

mattwigway:
Great - I will continue to push additional documentation to this branch - working on the skims now. @bstabler do you know the units of DRV_COM_WLK_DDIST__AM (and other DDIST variables) - is it in hundredths of a mile? Also, the travel times (e.g. WLK_COM_WLK_TOTIVT) in seconds? The values seem to match up, but they get divided by 100 in a lot of places which makes me think they might be in hundredths of minutes.

bstabler:
The TM1 transit skims come from Cube TRNBUILD, which outputs scaled int * 100 skims.  You can see the original Asim expressions [here](https://github.com/ActivitySim/activitysim/blob/master/activitysim/examples/example_mtc/configs/tour_mode_choice.csv), which divide by 100.

bstabler:
@mattwigway - I just realized that I think I closed this PR when I deleted develop2?  Can you create the PR to develop and then I'll go ahead and pull it?  Sorry about that.  Thanks.

PR #394, opened 2021-03-16
Telecommute

bstabler:
Eventually we will merge the telecommute models into the develop branch.  SEMCOG is still working on them


guyrousseau:
Will any consideration be given to the treatment of work-from-home / telecommuting in mode choice? 

image

bstabler:
we can close this since example_semcog on develop now has the example

Issue #139, opened 2017-01-04
auto post package to pypi


bstabler:
Currently the ActivitySim package is manually posted to https://pypi.python.org/pypi/activitysim.  We should automate this process.

bstabler:
We should also automatically set the version number in the docs/conf.py based on the version in setup.py

bstabler:
@jpn-- are we thinking of adding this to #440?

jpn--:
Yes, the new plan is to get at least this into #440.  I'm not yet sure if we can automate the entire release checklist within the budget for Phase 6b.

bstabler:
this is now assigned to me to finalize with the pypi credentials

bstabler:
this is now working

PR #427, opened 2021-06-14
pip package has some unnecessary trace files 

bstabler:
so increment version to be able to upload updated package


bstabler:
Eventually let's do #139 to prevent this from happening in the future

PR #447, opened 2021-07-02
feat: chunkless operation

jpn--:
It appears that I can't currently run a model with chunking just plain turned off unless it's in training mode, but training mode carries a lot of overhead monitoring memory usage.  If I set to production mode without an available chunk cache history file it falls back to training mode.

This is my crack at creating a fourth chunk training mode: 'disabled'.  (the first 3 are training, production, and adaptive). The idea for disabled mode is that we don't do chunking, and also don't check or log memory usage ever, so we can focus on performance assuming there is abundant RAM.

### DIFF
--- a/activitysim/core/chunk.py
+++ b/activitysim/core/chunk.py
@@ -579,8 +579,9 @@ def log_rss(trace_label, force=False):
     hwm_trace_label = f"{trace_label}.log_rss"
 
     if chunk_training_mode() == MODE_PRODUCTION:
-        trace_ticks = 0 if force else mem.MEM_TRACE_TICK_LEN
-        mem.trace_memory_info(hwm_trace_label, trace_ticks=trace_ticks)
+        # FIXME - this trace_memory_info call slows things down a lot so it is turned off for now
+        # trace_ticks = 0 if force else mem.MEM_TRACE_TICK_LEN
+        # mem.trace_memory_info(hwm_trace_label, trace_ticks=trace_ticks)
         return
 
     rss, uss = mem.trace_memory_info(hwm_trace_label)
@@ -593,10 +594,10 @@ def log_rss(trace_label, force=False):
 
 def log_df(trace_label, table_name, df):
 
-    assert len(CHUNK_LEDGERS) > 0, f"log_df called without current chunker."
-
     if chunk_training_mode() in (MODE_PRODUCTION, MODE_CHUNKLESS):
         return
+
+    assert len(CHUNK_LEDGERS) > 0, f"log_df called without current chunker."
 
     op = 'del' if df is None else 'add'
     hwm_trace_label = f"{trace_label}.{op}.{table_name}"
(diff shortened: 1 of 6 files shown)


jpn--:
@toliwaga it certainly would be helpful for you to review this, as you're the expert on chunk management.

toliwaga:
I'm out of the office on vacation until the beginning of August

jpn--:
Next steps: @esanchez01 and @stefancoe will test "disabled" mode to confirm it works correctly.

esanchez01:
@jpn-- , After testing this on 100k households using the SANDAG 1 Zone setup (multiprocessed), there are still memory CSV files in the output/log directory as well as chunk logging across various logfiles. This does not conform with the behavior of the 'disabled' mode as described in the original post:

> The idea for disabled mode is that we don't do chunking, and also don't check or log memory usage ever, ..

It seems like memory file creation is currently hard coded [here](https://github.com/jpn--/activitysim/blob/def96d6a31557abe8551f0c4eb19c6e3f7f95395/activitysim/core/mem.py#L193) in mem.py so changes may need to extend beyond just chunk.py. As for the chunk logging, it seems like empty logging is produced due to the placement of the 'disabled' chunk mode conditions in certain methods such as [here](https://github.com/jpn--/activitysim/blob/def96d6a31557abe8551f0c4eb19c6e3f7f95395/activitysim/core/chunk.py#L323) or from functions that that don't set chunk mode conditions such as [here](https://github.com/jpn--/activitysim/blob/def96d6a31557abe8551f0c4eb19c6e3f7f95395/activitysim/core/chunk.py#L929).

You can find the configs and log files for this test run [here](https://github.com/esanchez01/activitysim/tree/sandag-chunk-test/activitysim/examples/example_sandag/chunk_tests/sandag_1_zone_chunkless).

stefancoe:
I ran the model using 3 different modes using the same configuration of 20k households and mp turned off. The attached results show that the old code is fastest, 'disabled' is second but considerably slower and 'production' mode is the slowest. 


[chunkless_comparison.xlsx](https://github.com/ActivitySim/activitysim/files/6795544/chunkless_comparison.xlsx)

bstabler:
@esanchez01 to test again since this feature would be good to get working soon

esanchez01:
I retested this feature with the new commits. There are still memory CSV files and chunk/memory logging that are produced. However, the extent of the memory tracking and logging appears to have been reduced. The run time for this new run was about 65 minutes compared to 140 minutes for the previous test, so it seems like unnecessary memory/chunk calls were further avoided.

You can find the configs and log files for the new test run [here](https://github.com/esanchez01/activitysim/tree/sandag-chunk-test/activitysim/examples/example_sandag/chunk_tests/sandag_1_zone_chunkless_test_2).

bstabler:
@jpn-- please PR to develop when ready

bstabler:
@jpn-- , actually, please move this to a branch within this repo so we can work on it here.  Thanks.

jpn--:
This is now a branch in the main repo, per @bstabler request.  Closing this PR in favor of that.

PR #448, opened 2021-07-02
update example_semcog

albabnoor:
remove unused files and update configs to work with latest activitysim version.


bstabler:
@albabnoor - please PR to https://github.com/ActivitySim/activitysim/tree/example-semcog instead so we can park this on the branch for now.  Thanks.

albabnoor:
see #453

PR #453, opened 2021-07-09
semcog example

albabnoor:
update example_semcog


bstabler:
I accepted this into the example semcog branch but it still needs some input data for the university model extensions to run successfully

PR #440, opened 2021-06-21
Improved distribution system for conda-forge and pip

jpn--:
This branch contains work on #424

PR #455, opened 2021-07-13
Improved distribution system for conda-forge and pip

jpn--:
Continues work from #440 on distribution system.


bstabler:
@jpn-- generally looks good.  I have some specific comments:

- I think we should PR to develop so we can accept this without having to also release it (since master always has the released code).  I'd like to make some doc updates after we accept this as well.  If you agree, can you change this PR to develop?
- The new chunking command line argument is in GB as opposed to bytes like the actual setting.  I find this confusing.  I think they should be consistent.  
- gettingstarted.rst refers to python 3.8
- Please remove all the optional tags in HOW_TO_RELEASE.md.  Let's make the steps more straightforward and the optional items are just good to do anyway.
- HOW_TO_RELEASE.md should refer to doing PRs to develop (assuming we decide to PR to develop)
- development.rst Develop Install section refers to using pip to install/uninstall.  I think this should now be conda forge.
- development.rst Releases should now refer to HOW_TO_RELEASE.md

And a couple items to note:
- I think this should be v1.1.0 since we're releasing several improvements - build process, orca, pandas 1.3, and the faster maz to maz and chunking (after we do the merge on develop)
- Thanks for adding that conda is no longer free for organizations with more than 200 employees.  Are the miniforge instructions any different?

After the revisions and I accept, I'll test the process and make a few docs updates.  I also want to merge in a few of the other PRs.

jpn--:
@bstabler I like the idea of PR's to develop, and develop goes to master. I wrote up this change into development.rst an opened #461 to discuss/formally approve this plan.

jpn--:
re miniforge: it is functionally identical to Miniconda, the only difference being that the default channel is set to conda-forge's free channel vs. the Anaconda Inc. paid channel.

PR #471, opened 2021-08-18
Use Cython version of toolz

jamiecook:
Usage of activitysim within popultionsim fails with missing module cytoolz - i believe that activitysim is using cytoolz internally rather than toolz. 

Referenced by @bstabler in https://github.com/ActivitySim/populationsim/pull/139


jamiecook:
How on earth does this decrease coverage by 6%???

bstabler:
thanks

jamiecook:
no worries

Issue #477, opened 2021-08-26
Documentation for using estimation mode for model development


jpn--:
ActivitySim's estimation mode documentation needs to be expanded to better explain how to use Estimation Mode for model development, including:

- [ ] How `larch.ModelGroup` works to jointly estimate multiple models.
- [ ] How to manipulate the utility function of a model after it has been declared but before parameters are estimated. For advanced users, this can facilitate model development, allowing for estimating models that are not strict copies of an existing ActivitySim model spec.

bhargavasana:
Given updated documentation in #917, can this be closed?

PR #478, opened 2021-08-31
Fix notebooks

jpn--:
Work to address #477


bstabler:
@jpn-- is this done? thanks

jpn--:
This PR in its current state does not actually close any of the sub-items in #477, it's just a predicate bugfix to address problems that needed to be solved first.  The fixes herein can stand on their own and since we're doing a release soon, let's just merge this now.

PR #481, opened 2021-09-15
Contribute recent SANDAG updates to main repo

mxndrwgrdnr:
## I. Background

This PR is the result of recent development work during conversion of SANDAG's cross-border model from CTRAMP to ActivitySim. Changes that were too SANDAG-specific were moved into an `extensions/` directory, as can be seen in `examples/example_sandag_crossborder/extensions/`. The changes that remain were deemed to be relevant to the core ActivitySim project and generic enough to be included in this PR. None of the proposed changes contain references to SANDAG or the cross-border model.

Care was taken to ensure all changes submitted here are **fully backward compatible** with existing ActivitySim implementations. A new example that makes use of these new features was added to the `examples/` directory, with tests written against it. However the updated code base has also been tested against all existing examples and their accompanying tests, and currently **passes all tests**.

Most of the changes herein can be categorized into one of four groups: 1) new model steps; 2) new ActivitySim modules; 3) changes to existing models; and 4) changes to existing ActivitySim modules. The first two categories are the cleanest to merge, because they don't touch any pre-existing code. Changes in the latter two categories are more involved, and in most cases were made to either fix bugs or consolidate code.

### Brand new model steps:
- **[tour_scheduling_probabilistic](https://github.com/ActivitySim/activitysim/pull/481/files#diff-b941a3a8cd72760b09c2d4d8f8e2b58325b3c7e5bfe8e689d309f3166e0c0417)**
  - WHAT: Probabilistic alternative to logit-based tour scheduling 
  - WHY: Survey data is often too sparse to properly estimate a multinomial logistic regression model for tour scheduling. A probabilistic approach solves this problem by allowing the user to provide a probability lookup table instead of a logit specification.
  - HOW: This new step is modeled after the existing "trip_scheduling" model, which already uses a probability lookup table to assign departure times. Rather than replicating this code in a new model, the probabilistic assignment logic that the new steps have in common was consolidated into a new `abm/models/util/` module called [probabilistic_scheduling.py](https://github.com/ActivitySim/activitysim/pull/481/files#diff-a6785c231ab7035e5498d6c88ad134c5e9eb9d6d27348671da36f14f3567a5e8). This approach has the added benefit of streamlining the "trip_scheduling" code.
  - COMPATIBILITY: All new model steps are fully backwards compatible with existing ActivitySim implementations.
  - RECOMMENDATION: ActivitySim PR
- **[tour_od_choice](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9ef1ec1ce20706ed77f3e1a95ca59b2bcb5e559f4fd81b04033fc0dc68e18da8)**
  - WHAT: Jointly modeled tour origin and destination choice
  - WHY: Currently ActivitySim depends upon the existence of a hard-coded "home_zone_id" column in the households table. However, synthetic population data may not always have residential location assignments (e.g. a synthetic population based on aggregate travel demand). In fact, many potential ActivitySim users may not have access to a synthetic population at all. By allowing agents to choose a tour origin at the same time as they choose a tour destination, we can greatly relax the input data requirements and broaden the scope of analyses that are possible with ActivitySim.
  - HOW: This model works much like the other location choice models in ActivitySim, except the table of alternatives is comprised of origin-destination pairs rather than just destinations. The alternatives a created dynamically by merging the land use table against itself and allowing the user to specify a subset of zones to be used as tour origins.
  - COMPATIBILITY: All new model steps are fully backwards compatible with existing ActivitySim implementations.
  - RECOMMENDATION: ActivitySim PR

### Brand new modules:

- [`abm/models/util/tour_od.py`](https://github.com/SANDAG/activitysim/blob/develop/activitysim/abm/models/util/tour_od.py)
   - WHY: Much like the location_choice.py models ("workplace_location_choice" and "school_location_choice") rely on the tour_destination.py submodule to do their heavy lifting, tour_od.py helps to keep the "tour_od_choice" [model step definition code](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9ef1ec1ce20706ed77f3e1a95ca59b2bcb5e559f4fd81b04033fc0dc68e18da8) to a minimum.
   - COMPATIBILITY: The "tour_od.py" module is only used by the "tour_od_choice" model, so this change is completely backwards compatible will all existing ActivitySim implementations.
   - RECOMMENDATION: ActivitySim PR
- [`abm/models/util/probabilistic_scheduling.py`](https://github.com/ActivitySim/activitysim/pull/481/files#diff-a6785c231ab7035e5498d6c88ad134c5e9eb9d6d27348671da36f14f3567a5e8)
   - WHY: As explained above, this module consolidates the probabilistic departure time assignment that is common between the "trip_scheduling" model and the new "tour_scheduling_probabilistic" model.
   - COMPATIBILITY: Fully backward compatible after replacing [this](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9278809d36670ed13b751b9e2ef74910ab383cd1e148d5f5e998500e20a16903L175-L274) entire 100 line chunk of code with one call to the new module as seen [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9278809d36670ed13b751b9e2ef74910ab383cd1e148d5f5e998500e20a16903R232-R234).
   - RECOMMENDATION: ActivitySim PR

### Existing models with substantial changes:
- **[tour_mode_choice](https://github.com/ActivitySim/activitysim/pull/481/files#diff-3ef875279de811c19699e00cfeea22969f9f278ffe4afa19adca13753fbeb00d)**
   - WHAT: pseudo-trip generation and trip mode choice execution
   - WHY: Tour mode choice models are commonly specified to include trip-level mode choice logsum terms. This functionality is currently not supported by ActivitySim and is thus an obstacle for users wishing to convert existing travel demand models over to ActivitySim.
   - HOW: Use the new `trip.initialize_from_tours()` method to create a table of half-tour trips (one outbound, one inbound) based on the tour origin and destination of each tour for each potential tour mode choice alternative. Then use this table of pseudo-trips to call the "trip_mode_choice" model from within "tour_mode_choice". Save the trip_mode_choice logsums and append them to the tours table.
   - COMPATIBILITY: This new functionality gets triggered by a new tour_mode_choice model configuration [parameter](https://github.com/ActivitySim/activitysim/pull/481/files#diff-5501b7db5d4d3752e2d407c2b2514704aa34bdfbc0cb9be287835e3210ba9ad0R68). This parameter is completely optional (as seen [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-3ef875279de811c19699e00cfeea22969f9f278ffe4afa19adca13753fbeb00dR139)), making this change completely backwards compatible will all existing ActivitySim implementations.
   - RECOMMENDATION: ActivitySim PR
- **[stop_frequency](https://github.com/ActivitySim/activitysim/pull/481/files#diff-eb38659859a5bcc956b6e20de084d4c71c1e77dbd6d42bdf40628eba8543fc54)**
   - WHAT: As described above, this change moves the trips initialization code to `abm/models/util/trip.py`
   - WHY: Code legibility and extensibility. Trips can now be initialized by any model step. 
   - HOW: The stop_frequency.py module now just calls the `abm/models/util/trip.py` module [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-eb38659859a5bcc956b6e20de084d4c71c1e77dbd6d42bdf40628eba8543fc54R174). 
   - COMPATIBILITY: Fully backwards compatible.
   - RECOMMENDATION: ActivitySim PR
- **[trip_destination](https://github.com/ActivitySim/activitysim/pull/481/files#diff-217c88f844fe951db363021dd12dd27a9d6bcaf3cf8495f463d8afc82308e460)**
   - WHAT: As described above, trip OD instantiation is moved from [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-217c88f844fe951db363021dd12dd27a9d6bcaf3cf8495f463d8afc82308e460L924-L928) to [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9120a3e389f22867b275b480733262b4b228b541bc88d0994d1d88ddb466c785R220-R224).
   - WHY: Code legibility and extensibility. Trips can now be initialized by any model step.
   - HOW: The "trip_destination" model now only initializes trip ODs if called from within "trip_purpose_and_destination" as described in the comment [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-217c88f844fe951db363021dd12dd27a9d6bcaf3cf8495f463d8afc82308e460R938-R941).
   - COMPATIBILITY: Fully backward compatible.
   - RECOMMENDATION: ActivitySim PR
- **[trip_scheduling](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9278809d36670ed13b751b9e2ef74910ab383cd1e148d5f5e998500e20a16903)**
   - WHAT: 1) stop duration scheduling mode; and 2) inbound scheduling constraint updating
   - WHY: 1) When survey data is too sparse to populate a probability lookup table for all possible departure periods, stop duration probabilities are often used to simulate sequential trip departure times; 2) Under the existing trip scheduling process, it is possible for inbound trips to depart before the last outbound trip of a tour.
   - HOW: 1) A [new parameter](https://github.com/ActivitySim/activitysim/pull/481/files#diff-a4ee3ad510b3403bf7cc05a2fbb986263ec99d7313f58225671796b8298cc9c3R11) in the "trip_scheduling" .yaml config and an internal duration probability processing [method](https://github.com/ActivitySim/activitysim/pull/481/files#diff-a6785c231ab7035e5498d6c88ad134c5e9eb9d6d27348671da36f14f3567a5e8R118-R135) in the new "probabilistic_scheduling.py" module; 2) Update the `tour_earliest` constraint after the last outbound trip is scheduled as seen [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9278809d36670ed13b751b9e2ef74910ab383cd1e148d5f5e998500e20a16903R305-R307) using the new method defined [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9278809d36670ed13b751b9e2ef74910ab383cd1e148d5f5e998500e20a16903R107).
   - COMPATIBILITY: 1) The `scheduling_mode` parameter is completely optional, with the [default](https://github.com/SANDAG/activitysim/blob/develop/activitysim/abm/models/trip_scheduling.py#L219) set to the existing departure-based scheduling mode to ensure complete backwards compatibility; 2) This change is really a bug fix. So although it is fully backwards compatible with existing ActivitySim implementations, new inbound trip departure times are to be expected.
   - RECOMMENDATION: ActivitySim PR

### Existing modules with substantial changes:
- [`abm/models/util/trip.py`](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9120a3e389f22867b275b480733262b4b228b541bc88d0994d1d88ddb466c785)
    - WHY: Currently, the code to generate the ActivitySim "trips" table is spread across multiple modules. The [stop_frequency](https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/stop_frequency.py#L264-L268) step first instantiates the table, and then [trip_destination](https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/trip_destination.py#L926-L928) augments it with `origin` and `destination` columns that are inherited from the parent tour. The trip.py module provides a more intuitive home in the code base for this code, and consolidates it under a single module: [`initialize_from_tours()`](https://github.com/ActivitySim/activitysim/pull/481/files#diff-9120a3e389f22867b275b480733262b4b228b541bc88d0994d1d88ddb466c785R131-R242). By isolating the trips initialization process in this way, it also makes it possible to initialize trips from within any number of arbitrary model steps. This in turn allows the user to compute trip-level metrics (e.g. trip mode choice logsums) for use in a model step that precedes "stop_frequency".
    - COMPATIBILITY: As seen [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-eb38659859a5bcc956b6e20de084d4c71c1e77dbd6d42bdf40628eba8543fc54R174), the "trips" table will still be instantiated in the "stop_frequency" model, making this change fully backwards compatible with all existing ActivitySim implementations.
    - RECOMMENDATION: ActivitySim PR
- [`core/pathbuilder.py`](https://github.com/ActivitySim/activitysim/pull/481/files#diff-115915315914169e272a4031c29132f952b526e3722e8e55a9cf62f3762d78fe)
   - WHAT: Turn tracing off if certain conditions are met. 
   - WHY: When transit paths are filtered [here](https://github.com/activitysim/activitysim/blob/master/activitysim/core/pathbuilder.py#L197-L204) it is rare but possible that no viable transit paths remain. In this case we have to force tracing to be turned off or else a very convoluted series of errors will be generated.
   - HOW: Conditional clauses [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-115915315914169e272a4031c29132f952b526e3722e8e55a9cf62f3762d78feR154-R155), [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-115915315914169e272a4031c29132f952b526e3722e8e55a9cf62f3762d78feR654-R655), [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-115915315914169e272a4031c29132f952b526e3722e8e55a9cf62f3762d78feR659-R660), [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-115915315914169e272a4031c29132f952b526e3722e8e55a9cf62f3762d78feR672-R673), and [here](https://github.com/ActivitySim/activitysim/pull/481/files#diff-115915315914169e272a4031c29132f952b526e3722e8e55a9cf62f3762d78feR822-R823).
   - COMPATIBILITY: This is a bug fix and as such is fully backwards compatible with all existing ActivitySim implementations.
   - RECOMMENDATION: ActivitySim PR

## III. Contribution Review Criteria
| criterion | response |
|---|---|
|1. Does it contain all the required elements, including a runnable example, documentation, and tests? | Yes. |
|2. Does it implement good methods (i.e. is it consistent with good practices in travel modeling)? | Yes. |
|3. Are the runtimes reasonable and does it provide documentation justifying this claim? | Yes. Runtimes of existing ActivitySim implementations/examples are unaffected, and new models have runtimes comparable to existing models.|
|4. Does it include non-Python code, such as C/C++? If so, does it compile on any OS and are compilation instructions included? | No.|
|5. Is it licensed with the ActivitySim license that allows the code to be freely distributed and modified and includes attribution so that the provenance of the code can be tracked? Does it include an official release of ownership from the funding agency if applicable? | Yes, and n/a. |
|6. Does it appropriately interact with the data pipeline (i.e. it doesn’t create new ways of managing data)? | Yes.|
| 7. Does it include regression tests to enable checking that consistent results will be returned when updates are made to the framework? | Yes. |
| 8. Does it include sufficient test coverage and test data for existing and proposed features? | Yes, but according to Coveralls here this PR would currently decrease coverage by ~3%. |
| 9. Any other comments or suggestions for improving the developer experience? | I developed a pretty useful workflow for debugging model expressions using a combination of the IPython `%debug` magic command and some ActivitySim `pipeline` hacks. This approach allowed me to reload updated expressions files and re-run a model step without having to restart the ActivitySim session, which can take a long time even if taking advantage of model/skim caching. It would be a good idea to formalize some of these debugging techniques so that new developers need not re-invent the wheel each time.|


bstabler:
Our process for review is drafted in [#483](https://github.com/ActivitySim/activitysim/issues/483).  @mxndrwgrdnr - please link to the validation notebook in the comments above and add responses to the [Contribution Review Criteria](https://activitysim.github.io/activitysim/development.html#contribution-review-criteria).  Once done, @danielsclint will review.

mxndrwgrdnr:
@bstabler I've added my responses to the contribution criteria above, and the validation notebook can be found [here](https://nbviewer.jupyter.org/github/SANDAG/ABM/blob/5d41be0545f750edd1c08b91b09b254c1fff63c2/src/asim_cross_border/notebooks/cross_border_validation.ipynb).

activitysim/abm/models/tour_od_choice.py
```diff
@@ -0,0 +1,115 @@
...
+        tours,
+        persons,
+        households,
+        land_use,
+        network_los,
+        chunk_size,
+        trace_hh_id):
+
+    """
+    Given a set of previously generated tours, each tour needs to have an
```
danielsclint, line 33: Docstring needs to follow Sphinx-compatible format, so the documentation can be automatically built.

activitysim/abm/models/tour_scheduling_probabilistic.py
```diff
@@ -0,0 +1,100 @@
+# ActivitySim
```
danielsclint, line 1: Needs Sphinx-compatible docstrings for automated documentation.

activitysim/abm/models/accessibility.py
```diff
@@ -4,6 +4,7 @@
 
 import pandas as pd
 import numpy as np
+import os
```
danielsclint: This import is not needed.

activitysim/abm/models/accessibility.py
```diff
@@ -154,5 +155,7 @@ def compute_accessibility(land_use, accessibility, network_los, chunk_size, trac
 
     logger.info(f"{trace_label} computed accessibilities {accessibility_df.shape}")
 
+    logger.info("{trace_label} added {len(results.columns} columns")
```
danielsclint: Added XX columns to what? It might be helpful to append a table name to the logger so the user knows what has been modified. Also might consider this a debug-level statement rather than an info-level statement keep standard logs more trim.

activitysim/abm/models/initialize_los.py
```diff
@@ -112,7 +112,48 @@ def initialize_los(network_los):
                     np.copyto(data, np.nan)
 
 
-def compute_utilities_for_atttribute_tuple(network_los, scalar_attributes, data, chunk_size, trace_label):
+def initialize_tvpb_calc_row_size(choosers, network_los, trace_label):
```
danielsclint: Docstring needs to follow Sphinx-compatible format, so the documentation can be automatically built.

danielsclint: The method doesn't appear to be used anywhere in this pull request. Is this intended to be a part of the pull request? If so, where is the matching enhancement to take advantage of this new method?

activitysim/abm/models/initialize_tours.py
```diff
@@ -94,7 +94,11 @@ def initialize_tours(network_los, households, persons, trace_hh_id):
         model_settings=model_settings.get('annotate_tours'),
         trace_label=tracing.extend_trace_label(trace_label, 'annotate_tours'))
 
-    tours = patch_tour_ids(tours)
+    skip_patch_tour_ids = model_settings.get('skip_patch_tour_ids')
```
danielsclint: This settings flag needs to be documented.

activitysim/abm/models/tour_mode_choice.py
```diff
@@ -159,23 +226,24 @@ def tour_mode_choice_simulate(tours, persons_merged,
     # add cached tvpb_logsum tap choices for modes specified in tvpb_mode_path_types
     if network_los.zone_system == los.THREE_ZONE:
 
-        tvpb_mode_path_types = model_settings.get('tvpb_mode_path_types')
-        for mode, path_types in tvpb_mode_path_types.items():
+        tvpb_mode_path_types = model_settings.get('tvpb_mode_path_types', False)
```
danielsclint: Default value behavior should be updated in the documentation.

activitysim/abm/models/util/trip.py
```diff
@@ -123,3 +126,117 @@ def get_time_windows(residual, level):
...
     return np.concatenate(ranges, axis=1)
+
+
+def initialize_from_tours(tours, addtl_tour_cols_to_preserve=None):
+    """
+    Instantiates a trips table based on tour-level attributes: stop frequency,
+    tour origin, tour destination.
+    """
+
+    stop_frequency_alts = pd.read_csv(
```
danielsclint: Since we've committed to Orcas for dependency injection, the reading of the alts should probably happen in a method labeled annotated as injectable.

activitysim/abm/models/util/trip.py
```diff
@@ -123,3 +126,117 @@ def get_time_windows(residual, level):
...
+    2     32927         32927        0             work            1       False           2
+    3     32927         32927        0             work            2       False           2
+    4     33993         33993        1             univ            1        True           1
+    5     33993         33993        1             univ            1       False           2
+    6     33993         33993        1             univ            2       False           2
+
+    """
+
+    # previously in trip_destination.py
+    tour_destination = reindex(unique_tours.destination, trips.tour_temp_index).astype(np.int64)
```
danielsclint, line 224: Does this initialize every trip regardless of its trip_num with trip_origin=tour_origin and trip_destination=tour_destination?

activitysim/abm/models/util/trip.py
```diff
@@ -1,10 +1,13 @@
 # ActivitySim
 # See full license in LICENSE.txt.
 import logging
-
+import pandas as pd
 import numpy as np
 
-from activitysim.core.util import assign_in_place
+import numpy as np
```
danielsclint: Numpy is imported twice.

activitysim/abm/models/tour_mode_choice.py
```diff
@@ -106,9 +112,15 @@ def tour_mode_choice_simulate(tours, persons_merged,
...
         # TVPB constants can appear in expressions
-        constants.update(network_los.setting('TVPB_SETTINGS.tour_mode_choice.CONSTANTS'))
-
-    estimator = estimation.manager.begin_estimation('tour_mode_choice')
+        if model_settings.get('use_TVPB_constants', True):
+            constants.update(network_los.setting('TVPB_SETTINGS.tour_mode_choice.CONSTANTS'))
+
+    # don't create estimation data bundle if trip mode choice is being called
+    # from another model step (i.e. tour mode choice logsum creation)
+    if pipeline._PIPELINE.rng().step_name != 'tour_mode_choice_simulate':
```
danielsclint: _PIPELINE is coded as an "implied" protected method. It should not be referenced from other classes.

activitysim/abm/models/tour_mode_choice.py
```diff
@@ -106,9 +112,15 @@ def tour_mode_choice_simulate(tours, persons_merged,
         })
 
         # TVPB constants can appear in expressions
-        constants.update(network_los.setting('TVPB_SETTINGS.tour_mode_choice.CONSTANTS'))
-
-    estimator = estimation.manager.begin_estimation('tour_mode_choice')
+        if model_settings.get('use_TVPB_constants', True):
```
danielsclint, line 253: Is this constant documented?

activitysim/abm/models/tour_mode_choice.py
```diff
@@ -123,6 +135,61 @@ def tour_mode_choice_simulate(tours, persons_merged,
     primary_tours_merged['tour_purpose'] = \
         primary_tours_merged.tour_type.where(not_university, 'univ')
 
+    # if trip logsums are used, run trip mode choice
+    if model_settings.get('COMPUTE_TRIP_MODE_CHOICE_LOGSUMS', False):
+
+        # Construct table of hypothetical trips from tours for each potential
+        # tour mode. Two trips (1 inbound, 1 outbound) per [tour, mode] bundle.
+        # O/D, purpose, and departure times are inherited from tour.
+        primary_tours_merged['stop_frequency'] = '0out_0in'  # no intermediate stops
```
danielsclint: Is '0out_0in' nomenclature enforced by the ActivitySim software, or is '0out_0in' simply the conventional norm at this point?

activitysim/abm/models/tour_mode_choice.py
```diff
@@ -123,6 +135,61 @@ def tour_mode_choice_simulate(tours, persons_merged,
     primary_tours_merged['tour_purpose'] = \
         primary_tours_merged.tour_type.where(not_university, 'univ')
 
+    # if trip logsums are used, run trip mode choice
+    if model_settings.get('COMPUTE_TRIP_MODE_CHOICE_LOGSUMS', False):
```
danielsclint, line 278: Is this functionality documented?

danielsclint: In general, this code feels like it adds a lot of new functionality to this method. It would ease testing and understanding if this was introduced as a new standalone method called from this method.

activitysim/abm/models/tour_mode_choice.py
```diff
@@ -123,6 +135,61 @@ def tour_mode_choice_simulate(tours, persons_merged,
...
+
+        # Construct table of hypothetical trips from tours for each potential
+        # tour mode. Two trips (1 inbound, 1 outbound) per [tour, mode] bundle.
+        # O/D, purpose, and departure times are inherited from tour.
+        primary_tours_merged['stop_frequency'] = '0out_0in'  # no intermediate stops
+        primary_tours_merged['primary_purpose'] = primary_tours_merged['tour_purpose']
+        trips = trip.initialize_from_tours(primary_tours_merged)
+        trips['stop_frequency'] = '0out_0in'
```
danielsclint: If '0out_0in' is to be used, it should be set in a variable. Effectively, this is a "magic number"

activitysim/abm/models/tour_od_choice.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    trace_label = 'tour_od_choice'
+    model_settings_file_name = 'tour_od_choice.yaml'
+    model_settings = config.read_model_settings(model_settings_file_name)
+    origin_col_name = model_settings['ORIG_COL_NAME']
+    dest_col_name = model_settings['DEST_COL_NAME']
+    alt_id_col = tour_od.get_od_id_col(origin_col_name, dest_col_name)
+
+    sample_table_name = model_settings.get('OD_CHOICE_SAMPLE_TABLE_NAME')
```
danielsclint, line 67: These settings need to be documented.

activitysim/abm/models/trip_scheduling.py
```diff
@@ -300,56 +171,72 @@ def schedule_trips_in_leg(
...
+        ADJUST_NEXT_DEPART_COL = 'latest'
     trips.next_trip_id = trips.next_trip_id.where(~is_final, NO_TRIP_ID)
 
-    # iterate over outbound trips in ascending trip_num order, skipping the initial trip
-    # iterate over inbound trips in descending trip_num order, skipping the finial trip
     first_trip_in_leg = True
     for i in range(trips.trip_num.min(), trips.trip_num.max() + 1):
 
-        if outbound:
+        if outbound or scheduling_mode == 'stop_duration':
```
danielsclint: Consider a CONSTANT variable rather than a "magic string".

activitysim/abm/tables/skims.py
```diff
@@ -18,7 +18,7 @@
 
 
 @inject.injectable(cache=True)
-def network_los_preload():
+def network_los_preload():\
```
danielsclint: Remove line feed character.

activitysim/abm/models/util/tour_od.py
```diff
@@ -0,0 +1,980 @@
+# ActivitySim
+# See full license in LICENSE.txt.
+# Code here primarily based on tour_destination.py
+
```
danielsclint: Sphinx-based documentation needed.

activitysim/abm/models/util/tour_od.py
```diff
@@ -0,0 +1,980 @@
+# ActivitySim
```
danielsclint, line 1: Is there a reason this code is separated from activitysim.abm.models.tour_od_choice? Do other models use these utility functions?

activitysim/abm/models/util/tour_od.py
```diff
@@ -0,0 +1,980 @@
...
+    # 1       16421         2       13.0
+    # 2       24251         3       14.0
+
+    # print(f"TAZ_size_terms ({TAZ_size_terms.shape})\n{TAZ_size_terms}")
+    # print(f"MAZ_size_terms ({MAZ_size_terms.shape})\n{MAZ_size_terms}")
+
+    return MAZ_size_terms, TAZ_size_terms
+
+
+def choose_MAZ_for_TAZ(
```
danielsclint, line 275: Does this model only work with 2+ zone systems? I don't see any logic that would avert the need for MAZ and relying instead wholly on a one zone system.

activitysim/abm/models/trip_purpose.py
```diff
@@ -61,36 +63,40 @@ def choose_intermediate_trip_purpose(trips, probs_spec, estimator, trace_hh_id,
...
-        file_name = '%s.UNMATCHED_PROBS' % trace_label
-        logger.error("%s %s of %s intermediate trips could not be matched to probs based on join columns  %s" %
-                     (trace_label, len(unmatched_choosers), len(choosers), PROBS_JOIN_COLUMNS))
-        logger.info("Writing %s unmatched choosers to %s" % (len(unmatched_choosers), file_name,))
-        tracing.write_csv(unmatched_choosers, file_name=file_name, transpose=False)
-        raise RuntimeError("Some trips could not be matched to probs based on join columns %s." % PROBS_JOIN_COLUMNS)
-
-    # select the matching depart range (this should result on in exactly one chooser row per trip)
-    choosers = choosers[chooser_probs]
+    if use_depart_time:
```
danielsclint, line 70: What is the intended purpose of this filter?

activitysim/core/assign.py
```diff
@@ -183,9 +184,11 @@ def assign_variables(assignment_expressions, df, locals_dict, df_alias=None,
     lowercase variables starting with underscore are temp variables (e.g. _local_var)
     and not returned except in trace_results
 
-    uppercase variables starting with underscore are temp scalar variables (e.g. _LOCAL_SCALAR)
+    uppercase variables starting with underscore are temp singular variables (e.g. _LOCAL_SCALAR)
```
danielsclint, line 187: What is the reasoning behind changing the nomenclature?

activitysim/core/los.py
```diff
@@ -211,6 +211,7 @@ def load_skim_info(self):
         if self.zone_system == THREE_ZONE:
             # load this here rather than in load_data as it is required during multiprocessing to size TVPBCache
             self.tap_df = pd.read_csv(config.data_file_path(self.setting('tap'), mandatory=True)).sort_values('TAP')
+            self.tap_df = pd.read_csv(config.data_file_path(self.setting('tap'), mandatory=True))
```
danielsclint: This seems like a duplicate read. It should be either or, not both.

activitysim/core/skim_dictionary.py
```diff
@@ -361,7 +361,7 @@ def wrap_3d(self, orig_key, dest_key, dim3_key):
 class SkimWrapper(object):
 
     """
-    A SkimWrapper object is an access wrapper around a SkimDict of multiple skim objects,
+    A   ltiple skim objects,
```
danielsclint: Seems like an accidental delete.

activitysim/core/test/test_los.py
```diff
@@ -166,9 +166,9 @@ def test_30_minute_windows():
     assert network_los.skim_time_period_label(36) == 'PM'
     assert network_los.skim_time_period_label(46) == 'EV'
 
-    pd.testing.assert_series_equal(
+    np.testing.assert_array_equal(
```
danielsclint, line 169: Why switch from pd to np testing classes? Performance?

danielsclint (changes requested):
I added a variety of comments to the code. Most are related to documentation with a handful related to code syntax and functionality.

I didn't run the tests yet. I'll hold on that until the partners have a chance to understand the changes in the code and review the feedback from my review. 

At a minimum, I recommend a broader discussion with the partners about changes to the following components:
 - tour_mode_choice
 - tour_od_choice


Tour OD Choice
This model seems dependent on having an MAZ layer. If that is correct, are the partner comfortable deploying models that are only compatible with a 2+ zone system. In other words, should all new models support multiple zone systems?

activitysim/abm/models/accessibility.py
mxndrwgrdnr: Dropped this line altogether. I think it was a holdover from a previous version of the code that I re-merged in by mistake. There is no `results` object here at all.

activitysim/abm/models/tour_od_choice.py
mxndrwgrdnr: How's this?

activitysim/abm/models/tour_od_choice.py
mxndrwgrdnr: Do you mean I should describe them in the docstring?

activitysim/abm/models/tour_scheduling_probabilistic.py
mxndrwgrdnr: Done.

activitysim/abm/models/accessibility.py
mxndrwgrdnr: Removed.

activitysim/abm/models/initialize_los.py
mxndrwgrdnr: Deleted. I think this is a result of a messy rebase at some point.

activitysim/abm/models/initialize_tours.py
mxndrwgrdnr: Is there a standard approach ActivitySim uses for documenting settings that are read from .yaml? Or one that you suggest?

FWIW the `patch_tour_ids()` method itself is not documented and honestly I struggled to discern from the code itself why it was necessary.

Also I did change it just now to make the default evaluate to `False`, that way it will be backwards compatible with existing initialize_tours.yaml files and the user only needs to include this setting if they are aware of it.

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: No. Same question as before, though: how/where to document .yaml settings?

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: Fixed.

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: I think so? See here: https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/stop_frequency.py#L126-L127

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: Good idea. Will work on this.

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: If I'm understanding you correctly, this is fixed.

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: The `dict.get()` method has a built-in default which is `None`. I only changed this to `False` to make it a little more clear what the if statement below (L230) is doing. I could get rid of it if you'd prefer, and either leave L230 like it is (`None` will still evaluate to `False`) or change L230 to `if tvpb_mode_path_types is None:`. Any preference?

activitysim/abm/models/trip_purpose.py
mxndrwgrdnr: This makes the trip purpose model more flexible by allowing the user to specify whether or not the trip purpose probability lookups should be filtered by matching departure time ranges or not. Before, it was assumed that they should, and an error was thrown if the probability lookup table didn't have departure range columns.

activitysim/abm/models/trip_scheduling.py
mxndrwgrdnr: Done.

activitysim/abm/models/util/tour_od.py
mxndrwgrdnr: Yeah. From the initial PR comment:

> WHY: Much like the location_choice.py models ("workplace_location_choice" and "school_location_choice") rely on the tour_destination.py submodule to do their heavy lifting, tour_od.py helps to keep the "tour_od_choice" model step definition code to a minimum.

More than anything I just thought I was conforming to an ActivitySim design pattern. Could just as easily be combined IMO.

activitysim/abm/models/util/tour_od.py
mxndrwgrdnr: Fixed

activitysim/abm/models/util/tour_od.py
mxndrwgrdnr: I would say roughly 70-80% of this code is shared with one of [location_choice.py](https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/location_choice.py#L316), [tour_destination.py](https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/util/tour_destination.py#L209), or [trip_destination.py](https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/trip_destination.py#L162). So the short answer is that I don't have any reason to believe that this model is any less-compatible with one zone systems than those. That being said, I can't say with 100% certainty that it would work without testing it against a proper one zone model, which we haven't done here.

mxndrwgrdnr: Was there anything in particular that stood out to you here as potentially more problematic for one zone systems than those other models I mentioned?

activitysim/abm/models/util/trip.py
mxndrwgrdnr: fixed

activitysim/abm/models/util/trip.py
mxndrwgrdnr: Fixed. But for my own benefit, can you explain what this change achieves?

activitysim/abm/models/util/trip.py
mxndrwgrdnr: Correct, except only for outbound trips. Inbound trips uses tour_destination and tour_origin, respectively.

activitysim/abm/tables/skims.py
mxndrwgrdnr: Fixed.

activitysim/core/assign.py
mxndrwgrdnr: In translating CTRAMP UEC's to ActivitySim expressions, I encountered a number of instances where it was useful to be able to create and store complex objects (e.g. numpy matrices, ordered dictionaries, etc.) that do not vary across choosers or alternatives. I wanted to piggyback on the functionality of the _LOCAL_SCALAR syntax, but in doing so discovered that it already works for objects more complex than scalars. I changed the docstrings and internal method names to better reflect this capability.

activitysim/core/los.py
mxndrwgrdnr: Fixed

activitysim/core/skim_dictionary.py
mxndrwgrdnr: Whoops! Fixed.

activitysim/core/test/test_los.py
mxndrwgrdnr: The `skim_time_period_label()` method now returns an array as seen [here](https://github.com/ActivitySim/activitysim/pull/481/files/78763b6598c090889735a087239765df9b0963f3#diff-7c1c031529711ac262c8a56155be3b10a61a3f906b6c1ca38f5efcd927a5e499R580). Going to update the docstring for this method now to reflect the change.

mxndrwgrdnr:
Thanks for this @danielsclint. I believe I've responded to all of your comments/questions or implemented the suggested fix. 

One recurring question for me was regarding your requests to document certain model settings that are read from .yaml. I wasn't sure what the suggested method for doing this would be...in the docstrings of the methods that read the .yaml? In the .yaml itself?

Still to do for me is to move the trip mode choice logsum computation in tour_mode_choice.py into its own method.

activitysim/abm/models/tour_mode_choice.py
mxndrwgrdnr: Done. Broke it up into a few separate methods, incorporated under one larger one named `get_trip_mc_logsums_for_all_modes()`. Feedback welcomed.

mxndrwgrdnr:
OK @danielsclint, done with my initial responses and fixes. Would appreciate clarification on a few of your suggestions. Thanks.

mxndrwgrdnr:
Latest commit fixes #492

mxndrwgrdnr:
Latest commit fixes #494 for `settings_file_name` CLI arg

activitysim/abm/models/util/trip.py
danielsclint: Consistency across the platform in terms how variables are "injected". It doesn't change the overall functionality, but helps minimize different approaches to solving the same problem across the code. Effectively, make it easier to understand and future developers can assume the setup rather than having to check and make sure.

activitysim/abm/models/util/tour_od.py
danielsclint: Its the references to the maz system in general. If a MAZ system is not specified in the setup, what happens when the code attempts to developed a bridge between the TAZ and the seemingly null MAZ setup in a one zone system?

activitysim/abm/models/tour_mode_choice.py
danielsclint: Recommend "if tvpb_mode_path_types is None:"

danielsclint (commented):
Reviewed updates and changes. Overall looks good. One additional recommendation for the consortium moving forward.... Anytime a new flag is added to the settings.yaml or any model-specific yaml, that value should be documented. There are already several that are not documented. Eventually, we will have a whole secret menu that only the developers know exist. Moving forward, it's probably a good practice to have any new model parameters reviewed before they are accepted.

danielsclint:
I'm going to run tests now.

activitysim/abm/models/util/tour_od.py
mxndrwgrdnr: OK, that's what I thought. I think your question is a good one and I can't say that I have a definitive answer to it as we have not explicitly tested for this scenario. However, if you look at where the references to MAZ's appear, they should be restricted entirely to the scenario in which choice sets are constructive via "pre-sampling" as opposed to the standard sampling mechanism. And its true that pre-sampling only works for multi-zone systems, as the main difference between pre-sampling and regular sampling is that with pre-sampling the choice set is first constructed by sampling alts by TAZ, and then randomly assigning a MAZ within each TAZ based on size terms. In short, the MAZ references should only be encountered if the user has turned on pre-sampling, which they should only do if they are using a multi-zone system. And if you look at the other models I linked to in my previous comment, that's exactly how they work as well.

That being said, there are two places in the tour OD code where I can find a reference to the MAZ system outside of the pre-sampling code. The first is [here](https://github.com/ActivitySim/activitysim/blob/fa898d6b1df524602760a010453d023892520459/activitysim/abm/models/util/tour_od.py#L161-L162), and the other is [here](https://github.com/ActivitySim/activitysim/blob/fa898d6b1df524602760a010453d023892520459/activitysim/abm/models/util/tour_od.py#L920-L926).

If I cleaned up those two MAZ-references, would that satisfy your concern about this model working for one-zone systems?

danielsclint:
@mxndrwgrdnr: Please fix code style error causing tests to break, then its good to go.

mxndrwgrdnr:
Thanks @danielsclint. Should be fixed now. Had to fix one more bug I encountered once the tests got past the style error. 

Do you want to resolve the rest of conversations before making the merge? I can do it if you'd prefer.

bstabler:
we're almost done, @mxndrwgrdnr to add the cross border model example to Travis

PR #482, opened 2021-09-17
Resize SANDAG 1 and 2 Zone examples to match 3 zone example

esanchez01:
Currently, a SANDAG 1, 2, and 3 zone 'test' example exists but they are of different sizes (# households, # zones, etc.). @jpn-- suggested making these test examples of the same size as it would help with profiling and benchmarking across zone systems without needing to run the full examples every time.

This PR updates the 1 and 2 zone test examples so they include the same synthetic population and geographic extent as the 3 zone test example. The size of the 3 zone test example was slightly increased so that it covers the same TAZs as the 1 zone. All three of these test examples now include 90 TAZs and ~49K households.


jpn-- (approved):
This looks good, the examples all work.  Thanks!

Issue #499, opened 2021-10-12
chunking on linux


janzill:
On Ubuntu18.04 and probably other linux systems, chunking does not seem to be working correctly. When [adjusting memory chunks for sub tasks](https://github.com/ActivitySim/activitysim/blob/722d6e36b2210d5d24dfa2ac4a3504c1e8f75336/activitysim/core/mp_tasks.py#L735), the shared memory correction is [different for linux compared to win/mac](https://github.com/ActivitySim/activitysim/blob/722d6e36b2210d5d24dfa2ac4a3504c1e8f75336/activitysim/core/mem.py#L269). The name of the method suggests it is meant for use with rss calculations, however the default method is uss. For my local runs I simply set linux to behave like win/mac, however this is based on using hybrid_uss (or uss) for memory calculations, because then the chunking method as implemented for win/mac calculates linux's pss, which seems to be the best approximation to per-process memory usage.

If you aim to support rss on linux, then maybe the chunk size adaption method needs to be aware of which set size is used. However, using rss and hybrid_rss didn't work for me either with the current implementation and therefore I suggest simply treating all operating systems the same and removing the distinction.

A possible side effect of this could be that when using rss ActivitySim will work on operating systems other than win/mac/linux, which are explicitly excluded [here](https://github.com/ActivitySim/activitysim/blob/722d6e36b2210d5d24dfa2ac4a3504c1e8f75336/activitysim/core/mem.py#L282), but I haven't checked if there are any references to uss in that case - [psutils only returns uss for linux/mac/win](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info).

bstabler:
Let's close this issue and continue the discussion on #500

PR #500, opened 2021-10-12
Linux chunking

janzill:
Addresses #499


bstabler:
Thanks @janzill.  Have you tested this on Linux, Mac, and Windows?  I want to confirm before accepting this small PR.  Also, do you think we need any documentation (rst file) updates?  If so, can you add those to this PR?

janzill:
I have not tested this on windows or mac because there is no change in behaviour, only on linux for which there is a change.

I had a look at the documentation and I did not see a place where I could meaningfully add something, it is still valid as is and strikes a good balance between technical detail and being approachable I think.

Issue #502, opened 2021-10-13
example_mtc transit skims /100


bstabler:
As explained in #431, the transit skims in the example_mtc (and possibly other derivative examples) expressions have /100 in them, which is a hold over from the MTC skims, which are Cube TRANPLAN skims, which are scaled ints and therefore need to be /100 before use. These should be dropped if not needed.  @esanchez01 is going to look into fixing this by replacing the 100 with a global constant that the user can configure.

PR #505, opened 2021-10-19
Replace transit /100 with global constant

esanchez01:
For #502

Assumes MTC expressions containing an equivalent of '/100' (e.g. '/ 100', '/100.0') should be replaced with a user-defined global constant `TRANSIT_SCALE_FACTOR`.  Examples reliant on these expressions (example_mtc, example_multiple_zone, example_sandag) are now updated to use this constant as set in `constants.yaml`. It appeared that example_psrc also used '/100' in its expressions so it was updated as well. No other examples outside of those mentioned were updated.


bstabler:
this looks good @esanchez01, let me know when this is ready for review.

esanchez01:
@bstabler, with the latest commit this is now ready for review. Regress trips/tours need to be updated for the SANDAG examples but, as you mention during the technical call 10/19, it may be better to update them all at once after pulling in all the expression updates.

The latest commit prevents an overflow issue that was occurring during the accessibility step for the SANDAG examples. Since `TRANSIT_SCALE_FACTOR=1` for the SANDAG examples, instead of what was previously 100, [round trip transit times](https://github.com/esanchez01/activitysim/blob/master/activitysim/examples/example_mtc/configs/accessibility.csv#L29) could take on large negative values if no path is available. These negative values can cause overflow when evaluating the [decay function](https://github.com/esanchez01/activitysim/blob/master/activitysim/examples/example_mtc/configs/accessibility.csv#L31) and subsequently result in null values for transit accessibilities and eventually model failure. This is actually logged by the run:
`numpy: Warning: overflow encountered in exp expression: _decay = _rt_available * exp(_trPkTime * dispersion_parameter_transit)`
`numpy: Warning: overflow encountered in exp expression: _decay = _rt_available * exp(_trOpTime * dispersion_parameter_transit)`

The commit fixes this by setting the lower bound for transit times to 0.

bstabler:
Thanks @esanchez01.  Can you create a new branch within this repo and then PR this update to that branch?  Then we can accept this into the repo but not yet accept it to develop until all the example/expected test result updates are ready.

bstabler (commented):
see comments in the issue

bstabler:
@esanchez01 - I created a new branch for this and accepted it so I can easily run this and check test results.  Thanks.

PR #506, opened 2021-10-20
update xborder example configs and regression table

mxndrwgrdnr:


mxndrwgrdnr:
Closing this for now, as these changes to the xborder configs will first make their way to the SANDAG/ABM repo.

PR #534, opened 2022-01-10
Update LICENSE.txt

joecastiglione:
Updated license so that it is aligned with the consultant contract language assigning copyright to AMPO Research Foundation.


AMPO-Staff (approved):
AMPORF has reviewed and approved this request as it aligns with contractual obligations.

PR #539, opened 2022-02-22
pass through kwargs for location choice model estimation

### LINKED ISSUE #538: Unable to change default parameters In location choice estimation component models.
@jpn-- In the auto ownership component model, I can pass in parameters like the edb_directory like so:
`model, data = component_model(modelname, return_data=True, edb_directory = 'C:\output3\estimation_data_bundle\auto_ownership')`

But I don't think this is possible for the location choice models like school location because the parameters are not exposed in the component model functions:
https://github.com/ActivitySim/activitysim/blob/master/activitysim/estimation/larch/location_choice.py#L297

They are available in the location_choice_model function, so could be made available to each component model:
https://github.com/ActivitySim/activitysim/blob/master/activitysim/estimation/larch/location_choice.py#L35

jpn--:
Closes #538


stefancoe:
@jpn-- Any next steps for me?

stefancoe (approved):
Code looks good and works as expected. The auto_ownership model could use the same treatment- it allows specification of the data bundle dir, but some arguments are not available, e.g. the name of the coefficients file. Do you want to me to create an issue for this?

PR #571, opened 2022-06-07
Differentiate examples between quality and validity of example models

jpn--:
The "primary" example for ActivitySim has been called the "MTC" model.

This model draws its heritage from MTC's Travel Model 1, but over the years both our example and TM1 have evolved, so that they are no longer really the same model.  We've talked about renaming ours to make it clear that it's not MTC's model anymore.  We are grateful to MTC for donating their model as a starting point but they are no longer responsible for it themselves, so we propose to change the name. 

In this PR, we rebrand our primary example as a "prototype" MTC model, instead of calling simply an "example".  More generally, we propose a 3 tiered naming scheme for all examples:

- "production" examples are calibrated and validated by the relevant agency, and are intended to be replicas of an "official" travel model used by that agency, although generally we expect agencies to maintain independent repositories and not rely on the Consortium maintained version as a single source for the model.  At this time there are no production examples in the consortium's collection but we expect that to change soon.  (@JilanChen, is the SEMCOG model ready for this label? @wusun2 is the SANDAG Cross Border in this category?)
- "prototype" examples are not representative of any "official" travel model used by the relevant agency, but they are viewed as "ok" models by the consortium: they are at least loosely calibrated and/or validated, and typically contain at least some components or parameters specialized for the relevant region.  They may be in-development models that are not quite finished, or consortium maintained models that are derived from but now different from the official model of some region.  They should not be used in place of "official" models for policy analysis in any given region, but could serve as a donor model for new users who want to implement ActivitySim somewhere new (subject to all the caveats that go along with transferring models).  Current prototype models include MTC, ARC, MWCOG (open PR), Marin County(?)
- "placeholder" examples are computational testbeds that technically run but have not been calibrated nor validated in any meaningful way.  These examples are early stage development models used for testing purposes, and users are strongly cautioned not to use them for any policy or planning purpose.  Current placeholder models include PSRC, and the three iterations of the regional SANDAG model.


jfdman:
It isn't a totally fictional model; it is based on MTC TM1. IT could be called MTC_TM1_test or something to make it more clear what the lineage is?

joecastiglione:
I also have some concerns about adopting a fictional name.  The MTC model is a real model that has been used in (largely) it's current form for multiple RTP cycles.  This is important for potential users to know.  I would be in favor of something like Joel proposes, or maybe MTC_TM1.5, MTC_TM1.6, etc

jpn--:
It is my understanding that the thing labeled as "example_mtc" in ActivitySim is not MTC's model (version 1.5 or otherwise) and is different enough that it should not be simply adopted as proxy for an MTC model in policy analysis.  

@lmz, @FlaviaTsang, as keepers of the "real" MTC models, your thoughts here would be valuable.

bricegnichols:
Would using a detached (fictional) model make it easier to bring best practices learned during other implementations? I imagine that when the base MTC example was implemented, there were changes required that would be useful for other downstream users. Is there a reluctance to "edit" the base MTC model versus creating an updated best practice Gondor model? 

As an example, PSRC noticed that initial steps annotating households in the MTC example created various issues when using our survey data in estimation mode. Other implementations like SEMCOG's seemed to refine that step, but it didn't make it into the MTC example so we spent a while dealing with it ourselves. Rather than combing through other implementations based on the MTC model, could a Gondor model help consolidate improvements as the basic working version? I'm mostly thinking of building some continuity across implementations specifically for model estimation, which is a more opaque process for us at this point in terms of the process and troubleshooting resources. Perhaps this is better addressed elsewhere, but we did want to mention it as we think through the intended use of this example.

PR #572, opened 2022-06-09
Adding MWCOG + small example dataset

AndrewTheTM:
This is the MWCOG example setup and small example dataset.


AndrewTheTM:
Fixed code style issues in script causing failed test

jpn--:
Your tests were failing cause you opened this PR against `master` instead of `develop`.  The `master` branch currently can't pass tests because the latest version of numba triggers a deprecation warning from numpy.  This has been fixed in `develop`, but the fix is not yet released in `master`.

More generally, we don't allow PRs against `master`, only against `develop`, see [here](https://activitysim.github.io/activitysim/development.html#working-together-in-the-repository).  I moved it, and I expect the tests will pass in travis when run with the `develop` code.

AndrewTheTM:
Thanks @jpn-- , I wasn't sure if it was the same since it's data (now I know!)

AndrewTheTM:
@jpn-- Do I need to do something to cause the tests to re-run? The last test was before you moved this to develop.

PR #574, opened 2022-06-23
School Escorting

dhensle:
School drop-off and pick-up model implementation.  Initial design details were presented at the [March 8th meeting](https://github.com/ActivitySim/activitysim/wiki/Project-Meeting-2022.03.08).


JoeJimFlood:
@dhensle I have finished my review. Here are my comments:

It might be good to add a check that `NUM_ESCORTEES` and `NUM_CHAPERONES` as defined in the settings file are consistent with the alternatives in school_escorting_alts.csv.

[abm/models/school_escorting.py](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py)
- `determine_escorting_par[t]icipants()`
    - [Line 40-41](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L40): Should `model_settings.get()` be used to prevent KeyErrors?
    - [Line 47](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L47): Looks correct to me
    - [Line 49/59](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L49): I agree with the comment that the driving age/escorting age should be configurable, [especially if we want ActivitySim to be used outside of the US](https://en.wikipedia.org/wiki/List_of_minimum_driving_ages). If we do go in that direction should we have the mode choice spec files reflect that?
- `create_school_escorting_bundles_table()`:
    - [Line 215](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L215): Since the number of escortees is defined in the settings file, this should probably be `for child_num in range(1, NUM_ESCORTEES+1)` or something like that
    - [Line 225](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L225): Just putting 0 here and removing Line 204 would make the code easier to read and more consistent with the field definitions before and after.

[abm/models/non_mandatory_scheduling.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/non_mandatory_scheduling.py)
- `create_pure_school_escort_tours` and `assign_in_place` are imported but never called (the latter is imported twice)

[abm/models/non_mandatory_destination.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/non_mandatory_destination.py)
- `annotate` is imported despite not being used

[abm/models/non_mandatory_tour_frequency.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/non_mandatory_tour_frequency.py)
- Is Line 7 needed given that Line 22 exists? No other functions from `school_escort_tours_trips` are called

[abm/models/trip_purpose.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/trip_purpose.py)
- The function imported in Line 4 is not used

dhensle:
@JoeJimFlood Thank you for your great comments.  I have responded to each in **bold** below.  Can you please review the changes and ensure they are satisfactory?  Thanks!

It might be good to add a check that `NUM_ESCORTEES` and `NUM_CHAPERONES` as defined in the settings file are consistent with the alternatives in school_escorting_alts.csv.
**Agreed, added a new check_alts_consistency method.**

[abm/models/school_escorting.py](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py)
- `determine_escorting_par[t]icipants()` **fixed name typo**
    - [Line 40-41](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L40): Should `model_settings.get()` be used to prevent KeyErrors? **yes, good catch**
    - [Line 47](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L47): Looks correct to me **Thanks for checking.  This was a comment to myself when developing.  Removed comment.**
    - [Line 49/59](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L49): I agree with the comment that the driving age/escorting age should be configurable, [especially if we want ActivitySim to be used outside of the US](https://en.wikipedia.org/wiki/List_of_minimum_driving_ages). If we do go in that direction should we have the mode choice spec files reflect that? **added ability to specify options in the model yaml**
- `create_school_escorting_bundles_table()`:
    - [Line 215](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L215): Since the number of escortees is defined in the settings file, this should probably be `for child_num in range(1, NUM_ESCORTEES+1)` or something like that **yup, missed that one.  Good catch.**
    - [Line 225](https://github.com/dhensle/activitysim/blob/c6d342bff2bce22f3f6702b020188749ed1b0328/activitysim/abm/models/school_escorting.py#L225): Just putting 0 here and removing Line 204 would make the code easier to read and more consistent with the field definitions before and after. **I believe we actually do need to initialize before this and not just use 0.  We do not want to overwrite what was set in the previous iterations of the loop.  Left as is.**

[abm/models/non_mandatory_scheduling.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/non_mandatory_scheduling.py)
- `create_pure_school_escort_tours` and `assign_in_place` are imported but never called (the latter is imported twice) **Removed unused / re-imported references.  assign_in_place is actually used on line 50, so I left that one but removed the second import.**

[abm/models/non_mandatory_destination.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/non_mandatory_destination.py)
- `annotate` is imported despite not being used **I see it used on line 119?**

[abm/models/non_mandatory_tour_frequency.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/non_mandatory_tour_frequency.py)
- Is Line 7 needed given that Line 22 exists? No other functions from `school_escort_tours_trips` are called **Nope, removed line 7.**

[abm/models/trip_purpose.py](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/trip_purpose.py)
- The function imported in Line 4 is not used  **Removed.  No idea where this came from.**

JoeJimFlood:
@dhensle Everything looks good to me. Thank you so much for your work on this!

activitysim/examples/prototype_mtc_extended/test/regress/final_trips.csv
```diff
@@ -1,80 +1,85 @@
...
+242323338,644476,386761,work,2,True,2,4,16,30290417,,,work,,9,WALK_LOC,0.5894708670927961
+242323341,644476,386761,work,1,False,1,16,4,30290417,,,home,,14,WALK,0.5926711189721502
+242323617,644477,386761,escort,1,True,1,20,16,30290452,644478,outbound,escort,,9,WALK_LOC,2.2266987144854578
+242323621,644477,386761,escort,1,False,1,16,20,30290452,,,home,,10,WALK_LOC,2.0602667118131186
+242323625,644477,386761,escort,1,True,1,20,16,30290453,,,escort,,15,SHARED2FREE,0.02563756984130738
+242323629,644477,386761,escort,1,False,1,16,20,30290453,644478,inbound,home,,21,SHARED2FREE,-0.6067369442221502
+242323665,644477,386761,shopping,1,True,1,24,16,30290458,,,shopping,,12,WALK,2.6242779649960877
+242323669,644477,386761,shopping,1,False,1,16,24,30290458,,,home,,15,WALK,2.1220846081768716
+242323977,644478,386761,school,1,True,1,20,16,30290497,644478,outbound,school,,9,WALK_LOC,1.5478828044209911
+242323981,644478,386761,school,1,False,1,16,20,30290497,644478,inbound,home,,21,WALK_LRF,3.7659908084629183
```
jpn--: I notice here that there appears to be an inconsistency between trip mode choice of chauffeur and escortee.  Trip 242323629 is chauffeur escorting a child home from school, trip 242323981 is the child being escorted.  Purpose, destination, schedule all line up as intended.  But the chauffeur is driving (SHARED2FREE) while the child is using transit (WALK_LRF).  Reviewing the various code changes, it looks like this discrepancy is not a bug in code that has been written, but just something that has not been locked down; I don't recall if this is an intentional or unintentional omission.

activitysim/abm/models/util/canonical_ids.py
```diff
@@ -54,6 +57,167 @@ def enumerate_tour_types(tour_flavors):
...
+def read_alts_file(file_name, set_index=None):
+    try:
+        alts = simulate.read_model_alts(file_name, set_index=set_index)
+    except RuntimeError:
+        logger.warning(f"Could not find file {file_name} to determine tour flavors.")
+        return pd.DataFrame()
+    return alts
+
+
+def read_spec_file(file_name, set_index=None):
```
jpn--: It looks like `read_alts_file` and `read_spec_file` are identical.  Shall we use only one?

dhensle: Yes, only one is needed.  And this was actually caught in the flexible trip and tour id's review.  Resolved when merging in changes from develop branch.

activitysim/examples/prototype_mtc_extended/test/regress/final_trips.csv
dhensle: Post-processing functionality has been added to force escortee tour and trip modes to match the chauffeur modes.  For tour modes, there is ambiguity as the tour mode for the outbound leg can be different for the inbound leg since the pickup/dropoff can happen on different chauffeur tours. (The current implementation just uses the last available chauffeur tour mode.  Maybe a mode heirarchy should be implemented? But this would have to change between modeling regions...) I am not sure it really matters considering the trip modes are made to match downstream.

activitysim/examples/prototype_mtc_extended/configs/school_escorting_inbound.csv
```diff
@@ -0,0 +1,175 @@
...
+util_chauf2_age_u35_pe,Chauffer 2 Age 35 years or younger - Pure escort,(ptype_chauf2 == 4) & ((chauf1 == 4) | (chauf2 == 4) | (chauf3 == 4)),coef_chauf_age_u35_pe
+util_chauf1_time_to_work_or_univ_rs,Chauffer 1 Auto time to work or university - Ride Share,time_mand_to_home1 * ((chauf1 == 1) | (chauf2 == 1) | (chauf3 == 1)),coef_chauf_time_to_work_or_univ_rs
+util_chauf2_time_to_work_or_univ_rs,Chauffer 2 Auto time to work or university - Ride Share,time_mand_to_home2 * ((chauf1 == 3) | (chauf2 == 3) | (chauf3 == 3)),coef_chauf_time_to_work_or_univ_rs
+util_chauf1_walk_dist_to_work_or_univ_rs,Chauffer 1 Walk dist to work or university - Ride Share,(dist_mand1_to_home < 3) & ((chauf1 == 1) | (chauf2 == 1) | (chauf3 == 1)),coef_chauf_walk_dist_to_work_or_univ_rs
+util_chauf2_walk_dist_to_work_or_univ_rs,Chauffer 2 Walk dist to work or university - Ride Share,(dist_mand2_to_home < 3) & ((chauf1 == 3) | (chauf2 == 3) | (chauf3 == 3)),coef_chauf_walk_dist_to_work_or_univ_rs
+util_chauf1_walk_dist_to_work_or_univ_pe,Chauffer 1 Walk dist to work or university - Pure Escort,(dist_home1_to_mand < 3) & ((chauf1 == 2) | (chauf2 == 2) | (chauf3 == 2)),coef_chauf_walk_dist_to_work_or_univ_pe
```
jpn--: There's an error in the escorting spec here. `dist_home1_to_mand` isn't a valid variable, it should be `dist_home_to_mand1`.  And again on the following line for "2".  When running in normal simulation mode the error is avoided and our test can pass this module, as the coefficient on this line happens to be set to 0 in the associated coefficients file.  But this will cause errors if run in estimation mode or with sharrow enabled, neither of which is currently tested for the school escorting model.

dhensle: Good catch.  Fixed.  It looks like the outbound and conditional outbound spec files did not have this error.

PR #577, opened 2022-08-03
Fixed random seed generator to generate unsigned integers instead of …

JoeJimFlood:
The value of `_MAX_SEED` in random.py is currently set to the maximum 32-bit *unsigned* integer, but the default data type for `np.random.RandomState().randint()` is at 32-bit *signed* integer. This was causing an error when the default random seed was set to be `None` and triggering the random seed generation.

To test this, I ran the SANDAG 1-zone examples twice and the 3-zone example once, all of which were set to use random seeds. All runs completed successfully. I compared the mode share from the two 1-zone runs. They weren't exactly the same but at most were within 0.3%, which is what I would expect if the random seed generator worked correctly.


jpn--:
@JoeJimFlood do you want to take a stab at writing a test to run in our continuous integration testing, that confirms this change works as expected?  I envision one easy way would be setting the seed to None on to an existing test example and running to check that the final results to NOT exactly match the "expected" results with a zero seed.

JoeJimFlood:
@jpn-- I've never done that before but I'd love to try. Would I just need to edit the settings.yaml file in the test subdirectory of one of the examples? For example in [this file](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/placeholder_sandag/test/configs_1_zone/settings.yaml)?

JoeJimFlood:
@jpn-- I had opened this pull request before Version 1.1 was released. Would it be better if I closed it and created a new one with the same changes applied to the v1.1 code?

JoeJimFlood:
I added a test of the random number generator to the tests in actions. To test that I ran it using a version of the software that has that test but not the random generator bug fix, meaning that the random seed generator test should fail due to the seed always being zero. The results of that are [here](https://github.com/JoeJimFlood/activitysim/actions/runs/3421475156), where the random seed test failed and all the others passed. I guess that would mean...
task_failed_successfully

activitysim/test/random_seed/configs/settings.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+cleanup_pipeline_after_run: True
+
+output_tables:
+  h5_store: False
+  action: include
+  prefix: final_
+  sort: True
+  tables:
+    - trips
```
dhensle: I think you need persons here instead of trips.

activitysim/test/random_seed/test_random_seed.py
```diff
@@ -0,0 +1,127 @@
...
+        try:
+            os.mkdir(test_path("output"))
+        except FileExistsError:
+            pass
+
+        subprocess.run(["coverage", "run", "-a", file_path] + run_args, check=True)
+
+        # Read in output to memory to compare later
+        outputs[name] = pd.read_csv(
+            os.path.join(test_path("output"), "final_persons.csv")
```
dhensle: Final persons is being checked for the test, but the above settings file is only outputting trips... Is the right settings.yanl file being read?  Not sure how this is working.

activitysim/test/random_seed/test_random_seed.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+
+    check_outputs("0-a", "0-b", True)
```
dhensle: I suggest adding a comment at the top of this block describing what the test is doing.  Something like "running prototype mtc model through workplace_location with 3 random seed settings, 0, 1, None, and undefined.  final_persons.csv is compared to ensure the same setting returns the same variables and a different setting returns something different."

activitysim/test/random_seed/test_random_seed.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def example_path(dirname):
+        resource = os.path.join("examples", "prototype_mtc", dirname)
+        return pkg_resources.resource_filename("activitysim", resource)
+
+    def test_path(dirname):
+        return os.path.join(os.path.dirname(__file__), dirname)
+
+    def create_rng_configs(rng_base_seed=None):
+        new_configs_dir = test_path("configs_random_seed_=_{}".format(rng_base_seed))
```
dhensle: Can we add these configs_random_seed_* to a .gitignore in the test folder so git won't pick up a bunch of files after running locally?

HOW_TO_RELEASE.md
```diff
@@ -5,9 +5,9 @@
 01. Check that the branch you intend to release is passing tests on Travis.
     If it's not passing there you should not release it.
 
-00. Start from a completely clean conda environment 
-    and git repository.  Assuming you have `conda` installed, you can do so 
-    by starting where ActivitySim is not yet cloned (e.g. in an empty 
+00. Start from a completely clean conda environment
+    and git repository.  Assuming you have `conda` installed, you can do so
+    by starting where ActivitySim is not yet cloned (e.g. in an empty
```
dhensle: Why does this HOW_TO_RELEASE.md file have all these changes?  These changes look outside the scope of this PR and I am wondering if they were even included on purpose.  Can we revert this file back to the develop branch and handle these changes elsewhere?

activitysim/test/random_seed/test_random_seed.py
JoeJimFlood: The test isn't reading this settings file. There's a function `create_rng_configs` that creates a settings file for a specified random seed by copying the settings file stored in the prototype_mtc example and editing it to use the specified random seed (as well as removing all of the model steps not needed for the test). Each run of ActivitySim in the test then calls the appropriate settings file. I can change it to copy this settings file or remove this configs folder to reduce confusion.

HOW_TO_RELEASE.md
JoeJimFlood: Thanks for catching this. It was not intentional. I have a couple of hypotheses on how it happened, but the correct file is now there.

activitysim/test/random_seed/test_random_seed.py
JoeJimFlood: I added your suggestion verbatim.

activitysim/test/random_seed/configs/settings.yaml
JoeJimFlood: I deleted this directory to reduce confusion

dhensle:
Looks good to me, thanks Joe!

Issue #545, opened 2022-03-16
Need OP_SKIMS in trip destination model / error in ARC example model


jpn--:
There is a line in the spec for the example ARC model marked with the comment:

> THIS IS WRONG ***** RIGHT HERE **** I CHANGED THE BELOW TO MAKE WORK MECHANICALLY

https://github.com/ActivitySim/activitysim/blob/master/activitysim/examples/example_arc/configs/trip_destination.csv#L16

It appears that whomever was building this file wanted to access three kinds of skim references:
1. od_skims = fixed-origin to stop-location
2. dp_skims = stop-location to fixed-primary-destination
3. op_skims = fixed-origin to fixed-primary-destination (as if there was no stop)

The third type is not used in the canonical MTC model for trip destination, and it seems it was never coded into the trip destination component for ActivitySim.  The current ARC example is just using the od_skims again as a placeholder for the missing op_skims to make it work "mechanically" (i.e. not crash) but it sure looks to me like the model spec file is wrong.  The fix requires:

- [x] Make op_skims available in the trip destination component (and by extension, opt_skims; we can probably skip po_ and pot_)
- [ ] Correct the trip destination spec file in prototype_arc 
- [ ] Verify outputs from prototype_arc model
- [ ] Update tests

Note to developers: Do not close this issue until @guyrousseau (or his designee) can confirm that ARC gets the fix in their agency-maintained model as well, and not just in the consortium's example model.  cc: @i-am-sijia

guyrousseau:
Thanks Jeff for bringing this up to our attention. Indeed the ARC ActivitySim example is merely a prototype model application at this point in time. That said, we at ARC will confer with @i-am-sijia at WSP to get this taken care of, in accordance with your aforementioned observations and suggestions. Hence for now, this issue shall remain open until properly addressed and resolved.

jfdman:
@dhensle please see this issue - I believe this is being addressed by the SANDAG "trip" (really stop) destination choice model - please advise.

jpn--:
At this time, the infrastructure to allow for op_skims has been merged into ActivitySim, but as yet there is no example model or test that implements the change.

i-am-sijia:
I updated and tested the ARC example model with the change, will create a PR to reflect that.

jfdman:
I see the updated ARC stop model is in the latest release, closing issue.

PR #579, opened 2022-08-06
[Phase 7] Sharrow -- 1 Zone

jpn--:
This PR supercedes #542, and is written with black enabled to simplify code review (relatively)


docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,73 @@
...
+parent repo as its remote, you can find the code in
+[PR #579](https://github.com/ActivitySim/activitysim/pull/579), accessible on the
+command line using the `gh` tool:
+
+```sh
+cd activitysim
+gh pr checkout 579  # use `gh auth login` first if needed
+cd ..
+```
+
```
dhensle, line 34: Here and in the installation instructions section it points to are both referencing a specific camsys activitysim fork and branch. The docs should be updated to reference the main activitysim repository.

activitysim/core/simulate.py
```diff
@@ -445,86 +456,138 @@ def eval_utilities(
...
+        # TODO: make this smarter
+        # trace_label.startswith("trip_mode_choice") # uses '.isin(I_RIDE_HAIL_MODES)'
+        # or
+        trace_label.startswith("joint_tour_composition")
+        or trace_label.startswith("stop_frequency.social")
+        or trace_label.startswith("stop_frequency.shopping")
+        or trace_label.startswith("stop_frequency.eatout")
+        # or trace_label.startswith("trip_destination")
+    ):
+        sharrow_enabled = False
```
dhensle, line 477: Why is Sharrow is not enabled for these specific models?  Is this still applicable?  If so, comments on why would be very helpful for future development.

activitysim/core/simulate.py
```diff
@@ -445,86 +456,138 @@ def eval_utilities(
     -------
 
     """
+    start_time = time.time()
+
+    # from .flow import apply_flow # need import here to make injectable available
+    # from . import inject
+    # skim_dataset = inject.get_injectable('skim_dataset')
```
dhensle, line 463: I think these comments can be removed.  Looks like apply_flow is imported below only if sharrow is enabled.

docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,73 @@
...
+Lastly, a comprehensive performance testing suite on the prototype MTC model
+(warning: this takes hours!)
+
+```sh
+activitysim workflow sharrow-contrast/mtc_comprehensive
+```
+
+All these performance tests assume you have sufficient RAM to run without chunking.
```
dhensle, line 73: I think a better user guide to settings in general.  Let's say a user has a typical modeling server (20 cores and 200GB of RAM or so) and wants to run our full prototype_mtc example.   We should give them some step by step instructions on what settings they would likely need to change:  what do they need to put for chunking, what are the sharrow options (and how does this relate to chunking), where do they change the number of hhs, processors, etc.  Maybe this should go in the gettingstarted.rst and then just point them here for a few more details about the sharrow options and to chunking for more chunking details, etc.

Also, it looks like there were a lot of changes in the configs for example models where string comparisons were removed. It is my understanding that this is because they slow down sharrow.  That should be mentioned here.  And more generally the best practices for writing expressions that can be accelerated well by sharrow.

I don't see any documentation on skim encoding. Is that always enabled when sharrow is turned on? Or is there a way to separate skim encoding from utility spec compiling?

activitysim/core/mp_tasks.py
```diff
@@ -1385,34 +1385,56 @@ def skip_phase(phase):
...
+                    kwargs=shared_data_buffers,
+                )
             )
-        )
-        t0 = tracing.print_elapsed_time("setup shared_data_buffers", t0)
-        mem.trace_memory_info("mp_setup_skims.completed")
+
+            t0 = tracing.print_elapsed_time("setup shared_data_buffers", t0)
+            mem.trace_memory_info("mp_setup_skims.completed")
+            tracing.log_runtime("mp_setup_skims", start_time=start_time, force=True)
```
dhensle, line 1437: This line and above are repeated for both sharrow and non-sharrow.  Can reduce code repetition by removing them from if/else and just setting after.

activitysim/core/skim_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+                    )
+                else:
+                    x = x.assign(time_period=self.time_map[time_key])
+                key = main_key
+            else:
+                raise KeyError(key)
+
+        result = self.dataset.iat(**x, _name=key)  # iat strips data encoding
+        # if 'digital_encoding' in self.dataset[key].attrs:
+        #     result = array_decode(result, self.dataset[key].attrs['digital_encoding'])
```
dhensle, line 263: Why is this commented out?  If we have an encoding, don't we still need to decode when performing a lookup?

activitysim/core/fast_mapping.py
```diff
@@ -0,0 +1,62 @@
...
+
+@nb.njit
+def _fast_map(fm, target, dtype=np.int32):
+    out = np.zeros(len(target), dtype=dtype)
+    for n in range(target.size):
+        out[n] = fm[target[n]]
+    return out
+
+
+class FastMapping:
```
dhensle, line 14: Need some documentation for this class.

activitysim/core/timetable.py
```diff
@@ -47,6 +49,122 @@
...
+@nb.njit
+def _fast_tour_available(
+    tdds,
+    tdd_footprints,
+    window_row_ids,
+    window_row_ix__mapper,
+    self_windows,
+):
+    """
+
```
dhensle, line 61: A description here would be helpful.

activitysim/core/timetable.py
```diff
@@ -47,6 +49,122 @@
...
+    return out
+
+
+@nb.njit
+def _available_run_length(
+    available,
+    before,
+    periods,
+    time_ix_mapper,
+):
```
dhensle, line 98: Description here too.

activitysim/core/timetable.py
```diff
@@ -47,6 +49,122 @@
...
+
+@nb.njit
+def _available_run_length_2(
+    window_row_id_values,
+    windows,
+    person_to_row,
+    before,
+    periods,
+    time_ix_mapper,
+):
```
dhensle, line 131: And description here.  Not immediately obvious what the difference is between this method and the previous one is.

activitysim/core/timetable.py
```diff
@@ -196,13 +314,15 @@ def __init__(self, windows_df, tdd_alts_df, table_name=None):
         self.checkpoint_df = None
 
         # series to map window row index value to window row's ordinal index
-        self.window_row_ix = pd.Series(
-            list(range(len(windows_df.index))), index=windows_df.index
+        from ..core.fast_mapping import FastMapping
```
dhensle, line 317: Can we move this import to the top of the file?

activitysim/core/timetable.py
```diff
@@ -317,22 +437,40 @@ def tour_available(self, window_row_ids, tdds):
...
-        x = tour_footprints + (windows << I_BIT_SHIFT)
-
-        available = ~np.isin(x, COLLISION_LIST).any(axis=1)
-        available = pd.Series(available, index=window_row_ids.index)
-
+        # assert len(window_row_ids) == len(tdds)
+        #
+        # # numpy array with one tdd_footprints_df row for tdds
+        # tour_footprints = self.tdd_footprints[tdds.values.astype(int)]
+        #
```
dhensle, line 461: Looks like these checks are now handled by construction in _fast_tour_available().  Is that correct?  If not, then seems like we would still need them?  If so, then I think we can remove this commented code.

activitysim/core/timetable.py
```diff
@@ -459,47 +597,31 @@ def adjacent_window_run_length(self, window_row_ids, periods, before):
...
-                first_unavailable = np.where(
-                    (1 - available) * mask, time_col_ix_map, num_cols
-                ).min(axis=1)
-                available_run_length = first_unavailable - time_col_ixs - 1
-
-            chunk.log_df(trace_label, "mask", mask)
-            chunk.log_df(trace_label, "first_unavailable", first_unavailable)
+            # available = (self.slice_windows_by_row_id(window_row_ids) != I_MIDDLE) * 1
+            # chunk.log_df(trace_label, "available", available)
+            #
```
dhensle, line 612: Seems like we can remove this commented code since it all now takes place in _available_run_length_2().

activitysim/examples/optimize_example_data.py
```diff
@@ -0,0 +1,41 @@
+import os
+
+import openmatrix
+import pandas as pd
+
+
+def patch_example_sandag_1_zone(example_dir):
+
```
dhensle, line 8: This is here due to the mismatch between skims and zones, right?  So if we fix the example data, we can get rid of this file?  Seems like a better solution to the problem is to just fix the example and get rid of this patch.

activitysim/core/memory_sidecar.py
```diff
@@ -0,0 +1,130 @@
...
+            now = time.time()
+            if now > last_flush + flush_interval:
+                stream.flush()
+                last_flush = now
+            if event == "STOP":
+                stream.flush()
+                break
+
+
+class MemorySidecar:
```
dhensle, line 109: Documentation on class needed.

activitysim/abm/models/location_choice.py
```diff
@@ -232,16 +238,17 @@ def aggregate_size_terms(dest_size_terms, network_los):
     # aggregate MAZ_size_terms to TAZ_size_terms
     #
 
+    # TODO:SHARROW FIX ME
```
dhensle, line 241: This fixme is going to be addressed in the two-zone sharrow PR?

activitysim/core/assign.py
```diff
@@ -317,7 +352,34 @@ def to_series(x):
 
             # FIXME should whitelist globals for security?
             globals_dict = {}
-            expr_values = to_series(eval(expression, globals_dict, _locals_dict))
+            try:
+                expr_values = to_series(eval(expression, globals_dict, _locals_dict))
+            except ValueError:
```
dhensle, line 357: Would it be useful to add some logging statement here about what expression failed and what the global and local dicts are?

activitysim/abm/models/util/vectorize_tour_scheduling.py
```diff
@@ -484,35 +474,46 @@ def tdd_interaction_dataset(
...
         )  # catch this before we slice on available
 
-        alt_tdd = alt_tdd[available]
+        alt_tdd = alt_tdd_.isel({dimname: available}).to_dataframe()
 
         chunk.log_df(trace_label, "alt_tdd", alt_tdd)
 
         # FIXME - don't need this any more after slicing
-        del alt_tdd[window_id_col]
+        # del alt_tdd[window_id_col]
```
dhensle, line 516: Can we clean up all these above commented calls to alt_tdd?

activitysim/abm/models/util/vectorize_tour_scheduling.py
```diff
@@ -484,35 +474,46 @@ def tdd_interaction_dataset(
...
         tour_ids = np.repeat(tours.index, len(alts.index))
         window_row_ids = np.repeat(tours[window_id_col], len(alts.index))
+        chunk.log_df(trace_label, "window_row_ids", window_row_ids)
 
         alt_tdd = alts.take(alts_ids)
 
         alt_tdd.index = tour_ids
-        alt_tdd[window_id_col] = window_row_ids
+
+        import xarray as xr
```
dhensle, line 483: Move import to top of file?

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,51 @@
+name: ActivitySim Docs
+
+on:
+  push:
+    branches: [ main, develop, sharrow-black ]
+  workflow_dispatch:
+
+jobs:
+  docbuild:
+    if: ${{ github.repository }} == "camsys/activitysim"
```
dhensle, line 10: Should clean up these references before merging.

activitysim/core/flow.py
```diff
@@ -0,0 +1,914 @@
...
+    try:
+        yield
+    except Exception:
+        logger.error(f"error in {tag} after {timedelta(seconds=time.time()-t0)} {tag2}")
+        raise
+    else:
+        logger.info(f"completed {tag} in {timedelta(seconds=time.time()-t0)} {tag2}")
+
+
+class TimeLogger:
```
dhensle, line 58: Need some documentation for this class.

activitysim/core/flow.py
```diff
@@ -0,0 +1,914 @@
...
+            time_periods = []
+            for t in time_periods_:
+                if t not in time_periods:
+                    time_periods.append(t)
+            if zarr_file:
+                logger.info(f"looking for zarr skims at {zarr_file}")
+            if zarr_file and os.path.exists(zarr_file):
+                # TODO: check if the OMX skims are modified more recently than
+                #       the cached ZARR versions; if so do not use the ZARR
+                # load skims from zarr.zip
```
dhensle, line 283: This looks like an important TODO.  In the case of running models with global feedback loops, we certainly want to avoid using the same skim cache for each iteration simply because the check for new OMX skims wasn't inplace.  Suggest adding this functionality before merging.

activitysim/core/flow.py
```diff
@@ -0,0 +1,914 @@
...
+                f"df.trip_period     -> odt_skims.time_period",
+                f"df._dest_col_name -> dot_skims.{odim}",
+                f"df._orig_col_name -> dot_skims.{ddim}",
+                f"df.trip_period     -> dot_skims.time_period",
+                f"df._dest_col_name -> dpt_skims.{odim}",
+                f"df._stop_col_name  -> dpt_skims.{ddim}",
+                f"df.trip_period     -> dpt_skims.time_period",
+                f"df._stop_col_name    -> pdt_skims.{odim}",
+                f"df._dest_col_name -> pdt_skims.{ddim}",
+                f"df.trip_period     -> pdt_skims.time_period",
```
dhensle, line 622: I think we need to eventually add skim values to the primary tour origin. (See this [issue](https://github.com/ActivitySim/activitysim/issues/545).)  Can we just add them in here now?

activitysim/core/interaction_sample.py
```diff
@@ -62,83 +62,95 @@ def make_sample_choices(
...
+    #
+    # chunk.log_df(trace_label, 'choices_array', choices_array)
+    # chunk.log_df(trace_label, 'choice_probs_array', choice_probs_array)
+    #
+    # del alts
+    # chunk.log_df(trace_label, 'alts', None)
+    # del cum_probs_array
+    # chunk.log_df(trace_label, 'cum_probs_array', None)
+    # del alt_probs_array
+    # chunk.log_df(trace_label, 'alt_probs_array', None)
```
dhensle, line 132: There are large chunks of commented code that appear obsolete after sharrow enhancements that I think can be removed in this file.

activitysim/core/interaction_sample_simulate.py
```diff
@@ -263,6 +265,16 @@ def _interaction_sample_simulate(
...
+            #     choosers.loc[zero_probs], how='left', rsuffix='_chooser'
+            # )
+            # _z_trace_rows = np.ones(len(_z_interaction_df), dtype=np.bool)
+            # _z_interaction_utilities, _z_trace_eval_results \
+            #     = interaction_simulate.eval_interaction_utilities(
+            #     spec, _z_interaction_df, locals_d, trace_label,
+            #     _z_trace_rows,
+            #     estimator=estimator, log_alt_losers=log_alt_losers)
             # FIXME this is kind of gnarly, but we force choice of first alt
             probs.loc[zero_probs, 0] = 1.0
```
dhensle, line 279: Seems like there should be some logging here that states this occurs.

activitysim/core/skim_dataset.py
```diff
@@ -0,0 +1,303 @@
+import logging
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+from sharrow import array_decode
```
dhensle, line 6: Currently all calls to array_decode are commented out.  If we truly do not need them, we can get rid of the import.

Additionally, array_decode and array_encode are missing documentation in sharrow.

activitysim/core/configuration.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    Cleans up pipeline after successful run.
+
+    This will clean up pipeline only after successful runs, by creating a
+    single-checkpoint pipeline file, and deleting any subprocess pipelines.
+    """
+
+    sharrow: Union[bool, str] = False
+    """
+    Set the sharrow operating mode.
```
dhensle, line 165: Add a reference to the using-sharrow documentation.

docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,73 @@
+
+# Using Sharrow
+
+This page will walk through an exercise of running a model with `sharrow`.
```
dhensle, line 4: Add a reference to the sharrow settings in configuration.rst.

conda-environments/activitysim-dev-base.yml
```diff
@@ -0,0 +1,64 @@
+# Environment for development
+# Does not install ActivitySim or Sharrow, you must do that yourself
```
dhensle, line 2: I think we need an overall readme in the conda-environments folder listing what each of these environments are used for.  They are growing!

activitysim/core/configuration.py
```diff
@@ -0,0 +1,291 @@
...
+    * `require` - Use sharrow optimizations, and raise an error if they fail
+      unexpectedly.  This is the preferred mode for running with sharrow
+      if performance is a concern.
+    * `test` - Run every relevant calculation using both sharrow and legacy
+      systems, and compare them to ensure the results match.  This is the slowest
+      mode of operation, but useful for development and debugging.
+    """
+
+
+class ZarrDigitalEncoding(PydanticBase):
```
dhensle, line 183: I don't fully understand how to use this setting... maybe add some examples in the description or something?

activitysim/core/flow.py
```diff
@@ -0,0 +1,914 @@
...
+                f"df._orig_col_name -> opt_skims.{odim}",
+                f"df._park_col_name -> opt_skims.{ddim}",
+                f"df.trip_period    -> opt_skims.time_period",
+                f"df._park_col_name -> pdt_skims.{odim}",
+                f"df._dest_col_name -> pdt_skims.{ddim}",
+                f"df.trip_period    -> pdt_skims.time_period",
+            ),
+        )
+    else:
+        return {}
```
dhensle, line 659: Does it make sense to return an empty dict here or should this throw an error instead?

activitysim/core/flow.py
```diff
@@ -0,0 +1,914 @@
...
+                f"df._dest_col_name -> size_terms.stoptaz",
+                f"df.purpose_index_num -> size_terms.purpose_index",
+            ),
+        )
+    return locals_d
+
+
+def apply_flow(
+    spec, choosers, locals_d=None, trace_label=None, required=False, interacts=None
+):
```
dhensle, line 873: This looks like the primary entry point from interaction_simulate.py and simulate.py.  Some higher level documentation on this function and how it is calling sharrow, etc. would be very helpful.

dhensle (changes requested):
A majority of the requested changes are around documentation. There's a lot of stuff added in here and I think we need to document well to mitigate any increase into the already steep learning curve of ActivitySim development or usage.

There are a number of comments and areas about the handling of maz level skims that I left under the assumption they will be addressed in the sharrow two and three zone work.

activitysim/core/los.py
```diff
@@ -584,11 +665,26 @@ def get_skim_dict(self, skim_tag):
...
-
-        assert (
-            skim_tag in self.skim_dicts
-        ), f"network_los.get_skim_dict: skim tag '{skim_tag}' not in skim_dicts"
-        return self.skim_dicts[skim_tag]
+        # TODO:SHARROW: TAZ and MAZ are the same
+        sharrow_enabled = config.setting("sharrow", False)
+        if sharrow_enabled and skim_tag in ("taz", "maz"):
+            skim_dataset = inject.get_injectable("skim_dataset")
+            from .skim_dataset import SkimDataset
```
i-am-sijia, line 672: Move this import to the beginning?

activitysim/core/los.py
```diff
@@ -445,7 +464,7 @@ def load_data(self):
                 == NOT_IN_SKIM_ZONE_ID
             ).any()
 
-    def create_skim_dict(self, skim_tag):
+    def create_skim_dict(self, skim_tag, _override_offset_int=None):
```
i-am-sijia, line 467: add `_override_offset_int` to docstring

activitysim/abm/models/location_choice.py
```diff
@@ -232,16 +238,17 @@ def aggregate_size_terms(dest_size_terms, network_los):
...
+    sharrow_enabled = config.setting("sharrow", False)
+    if sharrow_enabled:
+        skim_dataset = inject.get_injectable("skim_dataset")
+    else:
+        skim_dataset = None
```
i-am-sijia, line 246: `skim_dataset` does not seem to be used, why keeping them here? They don't seem to return an updated `network_los` either.

activitysim/examples/prototype_mtc/configs/settings.yaml
```diff
@@ -58,6 +60,8 @@ input_table_list:
     rename_columns:
       TAZ: zone_id  # person_id is the required index column
       COUNTY: county_id
+    recode_columns:
+      zone_id: zero-based
```
i-am-sijia, line 64: Is there documentation about the recode setting?

activitysim/examples/prototype_mtc/configs/settings.yaml
```diff
@@ -221,11 +225,25 @@ output_tables:
...
+    - tablename: land_use
+      decode_columns:
+        zone_id: land_use.zone_id
+    - tablename: households
+      decode_columns:
+        home_zone_id: land_use.zone_id
+    - tablename: persons
+      decode_columns:
+        home_zone_id: land_use.zone_id
+        school_zone_id: nonnegative | land_use.zone_id
```
i-am-sijia, line 237: Need documentation here about what are the acceptable syntaxes. If users need these settings, we should describe them in the documentation.

activitysim/workflows/sharrow-contrast/mtc_mini.yaml
```diff
@@ -0,0 +1,44 @@
...
+      legacy: True
+      reference: True
+      tag:
+      resume_after:
+      fast: False
+      compile_n_households: 5000
+      main_n_households: 100000
+      config_dirs: configs
+      data_dir: data
+      instrument: False
```
i-am-sijia, line 31: What is the definition of `instrument`, is it for benchmarking? Seems all the tests have `instrument: False`, has it been tested when set to True?

activitysim/core/configuration.py
```diff
@@ -0,0 +1,291 @@
...
+
+    If not provided, defaults to `input_store`
+    """
+
+    index_col: str = None
+    """table column to use for the index"""
+
+    rename_columns: dict[str, str] = None
+    """dictionary of column name mappings"""
+
```
i-am-sijia, line 31:
```suggestion

    recode_columns: dict[str, str] = None
    """dictionary of column name recoding"""
    
```

activitysim/core/interaction_simulate.py
```diff
@@ -351,66 +596,144 @@ def _interaction_simulate(
...
+                interaction_utilities[trace_rows],
+                tracing.extend_trace_label(trace_label, "interaction_utils"),
+                slicer="NONE",
+                transpose=False,
+            )
+
+    if sharrow_enabled == "test":
+        try:
+            if interaction_utilities_sh is not None:
+                np.testing.assert_allclose(
```
i-am-sijia, line 708: This part *seems* repetitive. In Sharrow test mode, you are checking if the utilities calculated by Sharrow is close enough to the utilities calculated non-Sharrow. I thought `eval_interaction_utilities()` is already doing that (with line 420-490).

i-am-sijia:
I have not gone through every file yet, especially ones in activitysim/workflows/. Will finish them and add more comments this week.

conda-environments/activitysim-dev.yml
```diff
@@ -1,11 +1,13 @@
 # Environment for development
-# usage: $ conda env create --file=activitysim-dev.yml -n ASIM-DEV-TMP
+# usage: $ mamba env create --file=activitysim-dev.yml -n ASIM-SH
 channels:
 - conda-forge
+- nodefaults
 dependencies:
```
i-am-sijia, line 6: I installed Sharrow following the developer install instruction. Have we discussed adding Sharrow as a dependency here? If so, how ready is it to be added? If not, what are the concerns?

activitysim/core/assign.py
```diff
@@ -317,7 +352,34 @@ def to_series(x):
...
+                #     except Exception as err:
+                #         logger.error(repr(err))
+                raise
+
+            if (
+                sharrow_enabled
+                and np.issubdtype(expr_values.dtype, np.floating)
+                and expr_values.dtype.itemsize < 4
+            ):
+                # promote to float32, sharrow is not compatible with float less than 32
```
i-am-sijia, line 381: > sharrow is not compatible with float less than 32

This is an important detail to note, is it documented or in an issue? Just curious, what is the reason that it has to be no less than 32? So the assigned columns created in choosers data frame will have to be int32, is that correct?

i-am-sijia (changes requested):

activitysim/core/simulate.py
jpn--: We've now got sharrow running for all model components, so this has been removed.

activitysim/core/simulate.py
jpn--: yes

activitysim/core/assign.py
jpn--: It's actually a numba thing, not sharrow in particular.  I've revised the comment to note this.  Also, it's only floats, not ints, that are limited to 32 bits and up, ints can be 1-byte no problems.  So the only numpy datatype that's off limits here is float16, which is used rarely outside of GPUs anyhow.

activitysim/core/configuration.py
jpn--: Made a much larger enhancement to this part of the docs.

docs/dev-guide/using-sharrow.md
jpn--: On config changes and performance:  There have been minimal config changes, mostly the poorly performing utility spec files have been left intact if they work correctly.  Optimizing each spec is certainly worthwhile but outside the scope of our current task.  A small handful of spec changes have been made to ensure general compatibility, mostly moving commands from utility spec to preprocessors, usually because they execute another python function (e.g. reindex).  See new docs:
https://camsys.github.io/activitysim/sharrow-haircut-feature/dev-guide/using-sharrow.html#performance-considerations

On skim encoding:  it's never on if sharrow is not on.  If sharrow is enabled, digital encoding is controlled from network_los.yaml.  See new docs:
https://camsys.github.io/activitysim/sharrow-haircut-feature/dev-guide/using-sharrow.html#digital-encoding

On better user guide to settings in general (esp how to set up chunking):  yes we need this, but not part of this task/PR.  Sharrow is very focused on no-chunking model performance so far, we've done a bit of work looking at chunking compatibility but that is out of scope for this task.

conda-environments/activitysim-dev-base.yml
jpn--: each conda-env file is now documented at the top of the file with a description.

activitysim/core/configuration.py
jpn--: https://camsys.github.io/activitysim/sharrow-haircut-feature/dev-guide/using-sharrow.html#digital-encoding

activitysim/core/flow.py
jpn--: a bunch of comments and docstrings have been added to the code, and some discussion in the `using-sharrow` docs

activitysim/core/los.py
jpn--: this non-global import avoids circular references being created.

activitysim/abm/models/location_choice.py
jpn--: removed.

activitysim/core/interaction_simulate.py
jpn--: agree, removed.

conda-environments/activitysim-dev.yml
jpn--: install instructions have been updated.

activitysim/core/los.py
jpn--: done

docs/dev-guide/using-sharrow.md
jpn--: install instructions updated.

jpn--:
Close in favor of #611

PR #581, opened 2022-08-11
Flexible Number of Tour & Trip IDs

dhensle:
This pull request removes the hard coded max number of trips and tours available in ActivitySim.  For details, see the presentation [here](https://github.com/ActivitySim/activitysim/wiki/Project-Meeting-2022.08.09) and scope [here](https://github.com/ActivitySim/activitysim/wiki/Phase-7a-Scope-of-Work#task-8-extend-activitySim-code-to-enable-more-flexible-number-of-tours-and-stops).


dhensle:
All of the changes present in this pull request are also implemented in the school escorting pull request.  The school escorting pull request also includes additional changes to this part of the code to add "tour flavors" for school escorting.

JoeJimFlood (changes requested):
Looks great! A couple comments on canonical_ids.py:
- The functions `read_alts_file()` and `read_spec_file()` are identical. Was this intentional?
- I know I'm being a bit nitpicky here, but the comment in Line 445 reflects the code before this change when `MAX_TRIPS_PER_LEG` was hard-coded. I'm concerned that this might confuse future developers. Possibly have it say `1st in = max_trips_per_leg + 1` or something like that?

dhensle:
Thanks for your comments Joe.   And please continue to be "nitpicky" as that is what we need in these reviews! I have removed the deuplicative read_spec_file() and left the read_alts_file() in place.  The comment has been updated.

PR #584, opened 2022-08-12
Fix numpy

jpn--:
Dependabot is warning of a security issue with numpy < 1.22.  To fix it we need to also upgrade to numba 0.56.  

This PR is currently erroring because numba 0.56 is [not on conda-forge yet](https://github.com/conda-forge/numba-feedstock/pull/94), it is waiting on a [llvmlite update](https://github.com/conda-forge/llvmlite-feedstock/pull/62).  

We can try this again when the dependencies are available.

### DIFF
--- a/conda-environments/activitysim-dev.yml
+++ b/conda-environments/activitysim-dev.yml
@@ -23,9 +23,9 @@ dependencies:
 - myst-parser  # allows markdown in sphinx
 - nbconvert
 - nbformat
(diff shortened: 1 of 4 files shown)


jpn--:
On further review, the community has deemed that there is [no actual security issue](https://github.com/numpy/numpy/issues/18993#issuecomment-1017712798) with numpy < 1.22.

Issue #598, opened 2022-09-06
prototype_sandag_crossborder is unstable


jpn--:
In the prototype_sandag_crossborder, the reproducibility is unstable dependent on the number of processes used for the run.  The current model only has a single regression test case, running the model for 10 households in multiprocess mode with 2 processes. Changing the number of processes (e.g. from 2 to 3) or disabling multiprocessing (implicitly going from 2 to 1) changes the outcome of the model, causing the test to fail.  The results are reproducible if the number of processes is held fixed at 2.

The design of ActivitySim is supposed to be stable over multiprocessing with any number of processes.  We have tests for other models (e.g. see prototype_mtc or placeholder_sandag) that run the same regression test against both single and multi-process and get identical results.

PR #599, opened 2022-09-06
Feature: Sharrow for 2 and 3 zones

jpn--:
Adds compatability for Sharrow in 2 and 3 zone systems.

Note: tests for sandag xborder model do not pass due to pre-existing problems, see #598


jpn--:
Close in favor of the sharrow-comprehensive PR #601

wusun2:
@dhensle , is the xborder model bug fixed?  Looks this is a problem when we plug in Sharrow?

PR #606, opened 2022-09-14
Adding fixed-origin to fixed-primary-destination skim wrapper

dhensle:
Fix for issue #545

PR #611, opened 2022-09-29
[Phase 7] Complete Sharrow

jpn--:
This PR supercedes all other sharrow PR's, and contains a complete and integrated implementation of all Phase 7 tasks.

Published (draft) documentation is here: https://camsys.github.io/activitysim/feature-sharrow-tt


docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,379 @@
...
+In legacy mode, temporary variables are useful but they can consume substantial
+memory, as the variable is computed and stored for every row in the entire dataframe
+before it can be used in other expressions.  In sharrow, temporary variables are
+allocated, used, and freed for each row seperately, so no extra memory is required.
+
+### Switchable Expressions
+
+As a general rule, it is best to write each utility expression in a manner that
+is cross-compatible between sharrow and legacy evaluation modes, even if that
+means transitioning a few expressions or parts of expressions into preprocessors.
```
dhensle, line 181: What is a "sharrow compatible expression"?  Is it just talking about the above discussion on how any references need to be included in the numba compile? If so, I think we need to say this.  It would also be helpful to add a common example or two here showing the syntax and differences.

docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,379 @@
...
+uses a completely different evaluation system routed through numba. The `@` prefix
+does not need to be stripped or added anywhere, it is simply ignored. The expression
+can reference other columns of the main dataframe directly or indirectly, so that
+either "income" or "df.income" is a valid reference to that column in the main
+dataframe.  You can also reference other variables, including skims, time tables,
+and constants defined in in the model settings file.  However, you *cannot* write
+any arbitrary Python code.  External functions are not allowed unless they have
+already been compiled with numba's `njit` wrapper.  Typically, unless specifically
+constructed to be allowed for a model component, cross-referencing, merging or
+reindexing against tables other than the main dataframe is not allowed.
```
dhensle: How can a user see what is available?  Is it possible for the user to add functions or objects?

docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,379 @@
...
+```
+
+## Model Design Requirements
+
+Activating the `sharrow` optimizations also requires using the new
+[`SkimDataset`](skim-datasets) interface for skims instead of the legacy
+[`SkimDict`](activitysim.core.skim_dictionary.SkimDict), and internally
+recoding zones into a zero-based contiguous indexing scheme.
+
+### Zero-based Recoding of Zones
```
dhensle, line 27: Is there some assert or check in the code that looks for this if zones are not zero-based and sharrow is on?  If not, would probably be good to add.

docs/dev-guide/skim-dataset.md
```diff
@@ -0,0 +1,21 @@
...
+
+- [`SkimDict`](activitysim.core.skim_dictionary.SkimDict), the legacy
+  framework, that stores all skim data in one large omnibus array, with
+  various offset lookups and tools to access values.
+- [`SkimDataset`](activitysim.core.skim_dataset.SkimDataset), an
+  [xarray.Dataset]() based framework, which mimics the
+  [`SkimDict`](activitysim.core.skim_dictionary.SkimDict) interface, and
+  adds a number of features optimized specifically for use with `sharrow`.
+  This framework is automatically used when sharrow is enabled, and there
+  is no user configuration to enable it separately.
```
dhensle, line 14: Can you add some extra detail here about how skims are shared between processes when running in multi-process mode in legacy vs sharrow.  In particular, legacy code creates the shared data buffers using the multiprocessing.rawarray object, but this is skipped in mp_tasks.py for sharrow (line 1411).  Instead a sharrow skim dataset is created. But it is unclear how this memory is shared across sub-processes.  Can you please add some explanation?

activitysim/core/flow.py
```diff
@@ -0,0 +1,802 @@
...
+
+
+@inject.injectable(cache=True)
+def skim_dataset_dict(skim_dataset):
+    from .skim_dataset import SkimDataset
+
+    return SkimDataset(skim_dataset)
+
+
+def skims_mapping(
```
dhensle, line 252: Is it still possible to perform generalized calls to the network_los object to get skim values. e.g. an expression like network_los.skim_dict().lookup(arbitrary_orig, arbitrary_dest, 'SOME_OMX_CORE')? Or does the skim mapping / encoding break this functionality?

activitysim/core/flow.py
```diff
@@ -0,0 +1,802 @@
...
+                "df.trip_period     -> odt_skims.time_period",
+                f"df._dest_col_name -> dot_skims.{odim}",
+                f"df._orig_col_name -> dot_skims.{ddim}",
+                "df.trip_period     -> dot_skims.time_period",
+                f"df._dest_col_name -> dpt_skims.{odim}",
+                f"df._stop_col_name  -> dpt_skims.{ddim}",
+                "df.trip_period     -> dpt_skims.time_period",
+                f"df._stop_col_name    -> pdt_skims.{odim}",
+                f"df._dest_col_name -> pdt_skims.{ddim}",
+                "df.trip_period     -> pdt_skims.time_period",
```
dhensle, line 385: Bringing up #606 again. Can we just add the additional skim wrapper here?

activitysim/core/interaction_sample.py
```diff
@@ -62,83 +60,95 @@ def make_sample_choices(
...
-    # (same layout as cum_probs_arr)
-    alt_probs_array = probs.values.flatten()
-    chunk.log_df(trace_label, "alt_probs_array", alt_probs_array)
+    # cum_probs_array = probs.values.cumsum(axis=1)
+    # chunk.log_df(trace_label, 'cum_probs_array', cum_probs_array)
+    #
+    # # alt probs in convenient layout to return prob of chose alternative
+    # # (same layout as cum_probs_arr)
+    # alt_probs_array = probs.values.flatten()
+    # chunk.log_df(trace_label, 'alt_probs_array', alt_probs_array)
```
dhensle: Is there a reason why this tracing is no longer included?  I think we should either keep it (preferred) or remove this comment block.

activitysim/core/interaction_sample.py
```diff
@@ -62,83 +60,95 @@ def make_sample_choices(
...
+    #
+    # chunk.log_df(trace_label, 'choices_array', choices_array)
+    # chunk.log_df(trace_label, 'choice_probs_array', choice_probs_array)
+    #
+    # del alts
+    # chunk.log_df(trace_label, 'alts', None)
+    # del cum_probs_array
+    # chunk.log_df(trace_label, 'cum_probs_array', None)
+    # del alt_probs_array
+    # chunk.log_df(trace_label, 'alt_probs_array', None)
```
dhensle: I think we can remove the above comment block.

activitysim/core/interaction_sample.py
```diff
@@ -251,76 +264,192 @@ def _interaction_sample(
...
+
+        (
+            interaction_utilities,
+            trace_eval_results,
+        ) = interaction_simulate.eval_interaction_utilities(
+            spec,
+            choosers,
+            locals_d,
+            trace_label,
+            trace_rows,
```
dhensle, line 223: Looks like the above commented section is no longer needed (should be removed).  But this means that this trace_rows is going to always be None for sharrow.  Does this mean we can't see the interaction utilities when running with sharrow? I didn't see any part of the documentation discussing changes in the tracing functionality when using sharrow.  If indeed tracing outputs change when using sharrow vs legacy, that should be described in the docs.

activitysim/core/interaction_simulate.py
```diff
@@ -57,200 +67,537 @@ def eval_interaction_utilities(
...
+    #     e = e.replace("tt.adjacent_window_after(", "tt_adjacent_window_after(tt, ")
+    #     e = e.replace("tt.previous_tour_ends(", "tt_previous_tour_ends(tt, ")
+    #     e = e.replace("tt.previous_tour_begins(", "tt_previous_tour_begins(tt, ")
+    #     e = e.replace("tt.remaining_periods_available(", "tt_remaining_periods_available(tt, ")
+    #     updated_exprs.append(e)
+    # exprs = updated_exprs
+    #
+    # from .flow import apply_flow
+    # from . import inject
+    # skim_dataset = inject.get_injectable('skim_dataset')
```
dhensle: Another comment block that can likely be deleted.

activitysim/core/interaction_simulate.py
```diff
@@ -57,200 +67,537 @@ def eval_interaction_utilities(
...
-        # add df for startswith('@') eval expressions
-        locals_d["df"] = df
+        # # add numba accelerated versions of timetable functions to locals_d
+        # locals_d.update(
+        #     tt_adjacent_window_before=timetable.tt_adjacent_window_before,
+        #     tt_adjacent_window_after=timetable.tt_adjacent_window_after,
+        #     tt_previous_tour_ends=timetable.tt_previous_tour_ends,
+        #     tt_previous_tour_begins=timetable.tt_previous_tour_begins,
+        #     tt_remaining_periods_available=timetable.tt_remaining_periods_available,
+        # )
```
dhensle: Looks like the sharrow versions are below.  Suggest removing comment block.

activitysim/core/timetable.py
```diff
@@ -317,22 +482,40 @@ def tour_available(self, window_row_ids, tdds):
...
+        # x = tour_footprints + (windows << I_BIT_SHIFT)
+        #
+        # available = ~np.isin(x, COLLISION_LIST).any(axis=1)
+        # if isinstance(window_row_ids, pd.Series):
+        #     available = pd.Series(available, index=window_row_ids.index)
+        # elif isinstance(window_row_ids, xr.DataArray):
+        #     available = xr.DataArray(available, dims=window_row_ids.dims, coords=window_row_ids.coords)
         return available
 
     def assign(self, window_row_ids, tdds):
```
dhensle: The school escorting model need to update the timetable object with school escorting tours. It does that by calling this method (as seen [here](https://github.com/dhensle/activitysim/blob/school_escorting/activitysim/abm/models/school_escorting.py#L495)). It looks to me like this should still work, but wanted to confirm and make sure you were aware of this call.

activitysim/core/timetable.py
```diff
@@ -459,47 +642,31 @@ def adjacent_window_run_length(self, window_row_ids, periods, before):
...
+            # available[:, -1] = 0
+            #
+            # available_run_length = _available_run_length(
+            #     available,
+            #     before,
+            #     periods.to_numpy(),
+            #     self.time_ix._mapper,
+            # )
+            #
+            # max_available_run_length = available_run_length.max()
```
dhensle: Remove old code.

activitysim/core/timetable.py
```diff
@@ -317,22 +482,40 @@ def tour_available(self, window_row_ids, tdds):
...
+        #
+        # # t0 = tracing.print_elapsed_time("slice_windows_by_row_id", t0, debug=True)
+        #
+        # x = tour_footprints + (windows << I_BIT_SHIFT)
+        #
+        # available = ~np.isin(x, COLLISION_LIST).any(axis=1)
+        # if isinstance(window_row_ids, pd.Series):
+        #     available = pd.Series(available, index=window_row_ids.index)
+        # elif isinstance(window_row_ids, xr.DataArray):
+        #     available = xr.DataArray(available, dims=window_row_ids.dims, coords=window_row_ids.coords)
```
dhensle: Can remove old code.

activitysim/cli/run.py
```diff
@@ -298,14 +370,21 @@ def run(args):
...
     return 0
 
 
 if __name__ == "__main__":
 
-    from activitysim import abm  # register injectables
+    from activitysim import abm  # register injectables  # noqa: F401
 
     parser = argparse.ArgumentParser()
     add_run_args(parser)
```
dhensle, line 392: Sorry if this is outside of the scope of this PR, but is the `parser.parse_args(["--sum", "7", "-1", "42"])` below used? Seems like this line can be deleted.

activitysim/core/cleaning.py
```diff
@@ -0,0 +1,66 @@
...
+        base_df = inject.get_table(tablename).to_frame()
+    except (KeyError, RuntimeError):
+        # the basis table is missing, do not
+        return False
+    if base_df.index.name and f"_original_{base_df.index.name}" in base_df:
+        return True
+    return False
+
+
+def recode_based_on_table(values, tablename):
```
dhensle, line 46: Is it valuable for the user to know that recoding happens?  Looks like this function and the one above get called in los.py when loading data and in the init of the MazSkimDict in skim_dictionary.py.  Neither of these places inform the user that any recoding is happening under the hood.  Suggest putting a logging statement in here or where these are called.

dhensle (commented):
Looks good! Major themes: some old code was left in as comments that I think should be cleaned up and some further documentation questions.

activitysim/examples/prototype_arc/configs/joint_tour_scheduling.yaml
```diff
@@ -2,6 +2,7 @@ LOGIT_TYPE: MNL
 
 SPEC: tour_scheduling_joint.csv
 COEFFICIENTS: tour_scheduling_joint_coeffs.csv
+sharrow_skip: True
```
i-am-sijia, line 5: Why is `sharrow_skip: True` for this component? It is set to True in arc joint tour and nm tour scheduling, and psrc nm tour scheduling, but not in others. Is there documentation on when we want to set sharrow_skip = True?

docs/dev-guide/using-sharrow.md
jpn--: The listing of what's available is (under this PR) generally no different from what's available before, so users can refer to the existing documentation --- which is in many cases inadequate and set to be updated in Phase 8.  Adding functions is possible by creating an extension module that defines new functions.  Documenting how to do so is out-of-scope for this PR.

docs/dev-guide/using-sharrow.md
jpn--: check added https://github.com/ActivitySim/activitysim/pull/611/commits/231e7a321d0deaa1bc280e1bf3ec53b5d560f979

activitysim/core/flow.py
jpn--: Yes, the `SkimDataset`/`DatasetWrapper` classes exactly mimic the `SkimDict`/`SkimWrapper` interface, and the network_los.get_skim_dict(...) automatically gives you a SkimDataset instead of a SkimDict when sharrow is enabled. This allows direct calls like `network_los.get_skim_dict(...).lookup(arbitrary_orig, arbitrary_dest, 'SOME_OMX_CORE')`, without the caller needing to even know if sharrow is enabled or not.

activitysim/core/timetable.py
jpn--: done

activitysim/core/cleaning.py
jpn--: ok

docs/dev-guide/skim-dataset.md
jpn--: Added

docs/dev-guide/using-sharrow.md
jpn--: done

activitysim/core/timetable.py
jpn--: I agree, it looks like it should still work.  It's fine to touch/modify the timetable object to block more times, as long as we maintain the structure and format so sharrow receives the array it expects.

activitysim/core/interaction_sample.py
jpn--: Have removed comment block.  This isn't tracing output, it's just _chunk_ tracing for chunking memory management, but the numba-enhanced `sample_choices_maker_preserve_ordering` goes from probs and rands directly to choices without allocating memory for intermediate arrays.

activitysim/core/interaction_sample.py
jpn--: Docs added.

activitysim/core/flow.py
jpn--: Yep, that is done now.

docs/dev-guide/using-sharrow.md
```diff
@@ -0,0 +1,413 @@
...
+
+```sh
+cd activitysim
+git switch sharrow-black
+cd ..
+```
+
+If your local Git repository is configured with only the main ActivitySim
+parent repo as its remote, you can find the code in
+[PR #579](https://github.com/ActivitySim/activitysim/pull/579), accessible on the
```
i-am-sijia: Update this to the new PR. When it's merged, remember to update this part again.

.gitignore
```diff
@@ -72,6 +72,7 @@ _test_est
 *_local/
 *_local.*
 
+**/__sharrowcache__
```
i-am-sijia, line 75:
```suggestion
**/__sharrowcache__
**/skims.zarr
```
In my test runs, skims.zarr is being picked up  by git for placeholder_sandag/output_1

i-am-sijia (changes requested):
Looks good to me! I only have three comments.
I tested the placeholder_sandag multi-zone model with Sharrow on my windows machine, it ran fine.

docs/dev-guide/using-sharrow.md
jpn--: Since I expect we will be transitioning from PR to merged into the develop branch and then to an upcoming release relatively promptly, I am just removing this section of instructions entirely; users will need no special instructions for how to access the code because it will just be in ActivitySim without any extra steps.

activitysim/examples/prototype_arc/configs/joint_tour_scheduling.yaml
jpn--: These models have the `sharrow_skip` option set to true so they will not use sharrow.  This is not because sharrow is incompatible -- if it is enabled, it does work, and the results are correct.  But rather, it is because they specs for these models include a LOT of string comparisons, and string comparisons are slow in numba.  

It is possible to rewrite the spec to factor away all or nearly all string comparisons and achieve good performance, in a way that is 100% backward compatible. This has been done for the prototype_mtc model.  But rewriting specs to enhance performance for other example models is explicitly outside the scope of this PR.

I have added a note to each YAML file where `sharrow_skip` is used to explain why.  I also added to the documentation under `Using Sharrow > Sharrow Compatability and Limitations > Performance Considerations` (where the string comparison performance limitation is already documented) that explains that `sharrow_skip` can be used in these cases to disable sharrow until the spec has been updated for better performance.

i-am-sijia (approved):

PR #612, opened 2022-10-03
Operational updates

jpn--:
A wide variety of updates have been made in the Sharrow PR, including core enhancements that change how ActivitySim work, and ancillary enhancements that add non-core features that are convenient for debugging and performance monitoring.

This PR isolates a limited number of non-core features so they can be add to version 1.1, allowing easier analysis and comparison against the sharrow changes coming for version 1.2:

- `memory_profile` option, which launches time-sampling memory monitoring in a separate subprocess.
- A more resilient `resume_after` that will fall back to restarting the model at the beginning if the selected checkpoint is not found (especially useful for setting the resume_after checkpoint as "_") 
- An `--ext` option for the CLI interface for `activitysim run` which allows the user to give the location of an extension package of add-ons for a model; this can obviate the need for the `simulate.py` script in most of the examples with add-on packages.


i-am-sijia (approved):
Looks good to me.

PR #614, opened 2022-10-04
Dev accessibilities

ghost:
Draft pull request for disaggregate accessibilities. Works with 1 and 2 zone models. Tested but some issues remain:
- Optional sampling and soft-merge methods using kmeans/naive bayes require sklearn package not part of testing suite. Disabled these imports for now. However, the pre-sampling method could/should be improved to use the weighted sampling method used in activitysim.
- difficulty testing multiprocess, it appears vehicle model does not pass mp test. Unsure if related or isolated issue.


activitysim/abm/models/disaggregate_accessibility.py
```diff
@@ -0,0 +1,660 @@
...
+            for x in klist
+        ]
+
+        # Create hhid
+        households[hhid] = households.index + 1
+        households["household_serial_no"] = households[hhid]
+
+        # Assign persons to households
+        rep = (
+            pd.DataFrame(util.named_product(hhid=households[hhid], index=persons.index))
```
i-am-sijia: Seems like the proto population does not allow households to have different household sizes nor different household member composition. Is that intended?

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
```diff
@@ -0,0 +1,159 @@
...
+    PROTO_HOUSEHOLDS:
+        index_col: proto_household_id
+        zone_col: home_zone_id
+        rename_columns:
+            zone_id: home_zone_id
+        VARIABLES:
+            hinccat1: [1, 2, 3, 4]          # Income categories
+            hworkers: 1                  # Household workers
+            veh: [0, 1, 2]               # Household vehicles
+            persons: 2                   # Two persons household
```
i-am-sijia: Can household size vary in the proto population? e.g. `persons: [1, 2, 3]`. And can households have different composition? e.g. type 1 household has one working adult and one school child, type 2 household has one working adult and one retiree. In applications, household size and composition could have impact on mode chocie logsum.

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
```diff
@@ -0,0 +1,159 @@
...
+        index_col: proto_person_id
+        VARIABLES:
+            pnum: [1, 2]                # Person number
+            military: 4                 # Not military
+            pstudent: 3                 # Not attending school
+            educ: 13                    # Bachelor's
+            grade: 0                    # Not attending
+            timeFactorWork: 1           # mean
+            timeFactorNonWork: 1        # mean
+        mapped_fields:
```
i-am-sijia: I think users would benefit from more detailed documentation on these settings in case they want to change them. E.g. when does it require users to define mappings for all vs some attributes; what is the minimum required attributes to create/map (seems like at least the ones used in the mode choice and location choice uec specs). I could be wrong but it seems there are some underlying rules like no need to define two workers in the proto pop.

activitysim/abm/tables/disaggregate_accessibility.py
```diff
@@ -0,0 +1,267 @@
...
+        accessibility_cols
+    ):
+        return
+
+    # Find the nearest zone (spatially) with accessibilities calculated
+    # Note that from here on the 'home_zone_id' is the matched name
+    if "nearest_accessibility_zone_id" not in persons_merged_df.columns:
+        persons_merged_df = find_nearest_accessibility_zone(
+            persons_merged_df, proto_accessibility_df
+        )
```
i-am-sijia, line 169: `method` arg is not passed into the function here. Users cannot define the method in settings. Judging from the code, `method` can be one of `{"centroid", "skims", ...}`. Is the implementation complete?

activitysim/abm/models/disaggregate_accessibility.py
```diff
@@ -0,0 +1,660 @@
...
+            sample_idx = (
+                self.land_use.sample(frac=1)
+                .reset_index()
+                .groupby("TAZ")[id_col]
+                .first()
+            )
+            sample_idx = sorted(sample_idx)
+        elif method and method.lower() == "kmeans":
+            # Performs a simple k-means clustering using centroid XY coordinates
+            centroids_df = pipeline.get_table("maz_centroids")
```
i-am-sijia, line 157: Seems `kmeans` will not work with one-zone system. Can we add `assert` before this?

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
```diff
@@ -0,0 +1,159 @@
...
+                tour_category:  # tour purpose category, mandatory/non-mandatory
+                    1: "mandatory"
+                    2: "non_mandatory"
+                    3: "non_mandatory"
+        filter_rows: # list any matching conditions as pandas expression
+            - ~((df.tour_type == "work") & (df.person_num == 1))
+        JOIN_ON:
+            person_num: pnum            # Specifies which person variable to join the tours on
+
+# Merge on variables
```
i-am-sijia, line 126: Based on the code, the user can also specify `asof` and `method`. It would be good to have some documentation on that without having user to look at the code, even though they might be less likely used.

activitysim/abm/models/disaggregate_accessibility.py
ghost: Added assertion for 2-zone systems. 1-zone could be implemented in future but may not be needed if aggressive sampling isn't typically requires for one-zone systems. This will require more work to ensure variable zoned centroid tables are loaded properly in tables/disaggregate_accessibility.py.

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
ghost: You're correct, this was created as specified in the 'Disaggregate Accessibilities Implementation Plan' (https://github.com/ActivitySim/activitysim/wiki/Project-Meeting-2022.03.22). 

You can change the household size, but you can't have multiple household sizes. E.g., you can't have a 2 person household and a 3 person household because persons must then be mapped to the households and this greatly complicates the generation procedure.

I agree that users may want this functionality, but is a bit out of scope. However, I think a simple solution for complicated cases would be to just let users to optionally create their own person/household/tour templates tables manually in a csv that gets read in. I just added this to the latest commit.

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
ghost: These are person attributes specified in the plan document. However, the timeFactorNonWork and timeFactorWork are not necessary to ActivitySim and have been removed. But will mention in documentation that the proto tables must include all fields necessary to run your models.

activitysim/abm/tables/disaggregate_accessibility.py
ghost: Added user config functionality for this parameter. 

"# How the nearest sampled zone is found: skims or centroid distance
NEAREST_METHOD: skims"

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
ghost: These were mainly for internal testing since the proto-pop should be designed to merge exactly, but left the functionality. However, to your point I added some user documentation for someone that might want to use it, e.g., continuous variables.

activitysim/examples/prototype_mtc_extended/configs/disaggregate_accessibility.yaml
i-am-sijia: Thanks. Where can I find the plan document?

i-am-sijia:
@nick-fournier-rsg  Thank you for addressing my comments. One new question - Should at least one 2-zone example model be updated to include disagg accessibility? Right now disagg accessibility is only tested within the 1-zone MTC extended model.

i-am-sijia:
Confirmed sandag 2-zone example ran on my end, thanks for adding it.

Issue #620, opened 2022-10-25
[Bug] Parking location choice not using landuse attributes from alternative parking zones


i-am-sijia:
## Describe the bug
The current parking location choice model uses landuse attributes (such as parking cost, parking capacity) from trip destination zones instead of from alternative parking zones. This is problematic as it results in the same parking cost for all alternatives. Code changes are required to bring in attributes from alternative parking zones for the model.

For example, the table below shows a snapshot of the parking location choosers dataframe from the MTC TM2 implementation using the ActivitySim/develop codes. You can see the values of parking attributes (column CJ-CS) are not changing by alternative parking zones (column B). In the correct implementation, the values of parking attributes should vary based on the alternative parking zones.

![image](https://user-images.githubusercontent.com/28812722/197870848-7cccb0db-d560-41e2-ae30-e999eb81a531.png)

### Triggering line of code
This is the line that's dropping columns from alternative zones (the comment is incorrect, we do need size terms in alternatives):

https://github.com/ActivitySim/activitysim/blob/7ead422df617384ea9cd113b6ff84a2297632053/activitysim/abm/models/parking_location_choice.py#L231-L233

This is the line that's dropping columns from destination sample zones:
https://github.com/ActivitySim/activitysim/blob/7ead422df617384ea9cd113b6ff84a2297632053/activitysim/abm/models/parking_location_choice.py#L174

Size terms are obtained from destination zones in the [example implementation](https://github.com/ActivitySim/activitysim/blob/develop/activitysim/examples/prototype_arc/configs/parking_location_choice_annotate_trips_preprocessor.csv#L5-L8).

### Thoughts on resolution
WSP has implemented the code changes to bring in attributes from alternative parking zones as part of the MTC ActivitySim implementation. Code changes are currently under review, and will be brought into SANDAG's ActivitySim implementation. WSP will create a pull request to the consortium repo once the review is done.

After this fix, the choosers dataframe looks like the screenshot below, the values of parking attributes vary based on the alternative parking zones.

![image](https://user-images.githubusercontent.com/28812722/197871014-89ffc849-3260-4eb4-abce-a956eaa15372.png)

@guyrousseau - @sidharthanr will update the ARC implementation to incorporate this fix.

guyrousseau:
[...]

PR #622, opened 2022-10-27
Parking location bug fix

i-am-sijia:
This PR is to address the bug in parking location choice documented in #620. 

The bug fix was done as part of the MTC TM2 ActivitySim implementation. It will be reviewed and adopted by SANDAG ActivitySim implementation. @kulshresthaa made the original code changes and his commit history can be found [here](https://github.com/SANDAG/activitysim/pull/46).

### DIFF
--- a/activitysim/abm/models/parking_location_choice.py
+++ b/activitysim/abm/models/parking_location_choice.py
@@ -297,7 +297,7 @@ def parking_location(
     land_use_df = land_use.to_frame()
 
     locals_dict = {"network_los": network_los}
-    
+
     constants = config.get_model_constants(model_settings)
 
     if constants is not None:
--- a/activitysim/abm/models/parking_location_choice.py
+++ b/activitysim/abm/models/parking_location_choice.py
@@ -171,7 +171,6 @@ def choose_parking_location(
     )
     destination_sample.index = np.repeat(trips.index.values, len(alternatives))
     destination_sample.index.name = trips.index.name
-    destination_sample = destination_sample[[alt_dest_col_name]].copy()
 
     # # - trip_destination_simulate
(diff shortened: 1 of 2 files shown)


jpn-- (commented):
So, this code change appears consistent with the intent and from my review of the results in the linked BayDAG repo, it appears to fix the problem.

But:  this PR changes the parking location choice model, without touching a single thing in the ARC example implementation -- no config.yaml changes, no spec.csv changes, and, most of all, no changes to any ARC test setup or regression test outcome.  And yet, all the tests still pass.  This suggests to me that (a) this code change is effectively not being tested adequately in this new PR, and (b) the parking location choice model is probably not being adequately exercised by any test prior to this PR either.

i-am-sijia:
> So, this code change appears consistent with the intent and from my review of the results in the linked BayDAG repo, it appears to fix the problem.
> 
> But: this PR changes the parking location choice model, without touching a single thing in the ARC example implementation -- no config.yaml changes, no spec.csv changes, and, most of all, no changes to any ARC test setup or regression test outcome. And yet, all the tests still pass. This suggests to me that (a) this code change is effectively not being tested adequately in this new PR, and (b) the parking location choice model is probably not being adequately exercised by any test prior to this PR either.

We did not need to update the parking location UEC for the actual ARC implementation, the code change fixed the bug. The parking location results after this update matches ARC's CTRAMP implementation, as Raghu noted in his follow-up comments in #620.

I took a look at the prototype ARC trip list, there are no trips having actual parking zone values - they all park at the destination zones - both before and after this change, which explains why the tests still pass. To your point, parking location choice model is not being adequately tested with the current prototype model.

PR #634, opened 2023-01-05
ARC Example Trip Destination UEC Correction

### LINKED ISSUE #545: Need OP_SKIMS in trip destination model / error in ARC example model

i-am-sijia:
This is to fix #545


jpn-- (dismissed):
LGTM

Issue #627, opened 2022-12-08
[Bug -- specification -- minor] Persons selecting `work_from_home` not excluded from telecommuting frequency


DavidOry:
## Describe the bug
As discussed in #609, a person "working from home" in ActivitySim does not have a usual work place and never commutes in the model. In the current [SEMCOG prototype](https://github.com/ActivitySim/activitysim/tree/main/activitysim/examples/prototype_semcog), persons selecting to work from home in the [work from home module](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/prototype_semcog/configs/settings.yaml#L85) are then subjected to the [telecommuting frequency module](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/prototype_semcog/configs/settings.yaml#L93). This is inconsistent with the ActivitySim definitions: persons working from home never telecommute because they do not have a usual work location. 

The implications of the bug should be minor. The primary impact of the telecommute frequency module is on the downstream coordinated daily activity pattern (CDAP) module. But those working from home are properly excluded from selecting a [mandatory pattern in CDAP](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/prototype_semcog/configs/cdap_indiv_and_hhsize1.csv#L65). Because those working from home have erroneously received a telecommuting frequency outcome, the [coefficients specific to these outcomes in CDAP](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/prototype_semcog/configs/cdap_indiv_and_hhsize1.csv#L61) will be misapplied, slightly changing the non-mandatory and home patterns for those working from home.  A very small amount of unnecessary computation is also being done due to the error. 

The main reason to fix the error is that it is confusing to users. As discussed in #609, the definitions related to telecommute and work from home are already challenging to keep straight. The model outcomes should reinforce these definitions, rather than contradict them.  

## To Reproduce
This error was discovered by our colleagues in Australia who are applying the work from home and telecommuting model to an ActivitySim prototype in Melbourne (thanks Andrea Colaiacomo). It can be reproduced by doing a cross tabulation of the telecommute frequency and work from home outcomes. 

### Failing tests
There are no tests for this error and probably do not need to be. 

### Thoughts on resolution
Adding @jfdman, @aletzdy, @dhensle, and @AndrewTheTM  to discuss resolution.

[...]

PR #635, opened 2023-01-27
Composite - Disaggregate Accessibility

jpn--:
This merged PR supersedes #614.  Several problematic interactions between disaggregate accessibility and shadow pricing have been resolved.


jpn--:
This merge also implements the workplace location fix suggested in #627, and includes everything from #574 as well.

Issue #640, opened 2023-02-02
maz_to_maz tables potentially cause duplicate index key error in los.py when using zero-based zone_id setting


stefancoe:
This error can happen [here](https://github.com/ActivitySim/activitysim/blob/main/activitysim/core/los.py#L298) when using MAZ IDs in the maz_to_maz tables that are higher than the maximum zero-based MAZ ID. This code is used to calculate the index of these files:
` df["i"] = df.OMAZ * self.maz_ceiling + df.DMAZ`

For example, our data has zone pairs 48-4903 & 55-997, with a max ceiling of 558, which makes df['i'] = 31687 in both cases. 

The reason our MAZ labels are so much higher than the zero-based maximum is because I am using a subset of our region for a testing dataset. However, this still happens in our full sized data presumably because zones without emp/hhs are removed in the [psrc_crop script](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/placeholder_psrc/scripts/psrc_crop.py). So we have IDs in maz_to_maz_walk file, for example, that are higher than the zero-based max.  

It seems like the maz_to_maz tables need to get the zero-based treatment as well?

jpn--:
Yeah, that's a bug.  When it's used, zero-based encoding should get applied immediately on load (as soon as possible after pd.read_csv) to any file that has MAZ or TAZ ids in it.  I will fix it.

stefancoe:
@jpn-- Thank you!

stefancoe:
@jpn-- Any update on this? Thanks!

jpn--:
Yes, I've found the fix in #643 is not robust across multiple use cases, including multiprocessing and large-value MAZ ID's.  The problem also interacts with / is related to #652.  I'm trying to get to a solution that works for all, but it looks like it will require duplicating the TAZ table, so one copy can be sliced for multiprocessing and the other not.

stefancoe:
Thanks! I actually meant to ask about #652. Any updates on that?

stefancoe:
@jpn-- Thank you!

stefancoe:
@jpn-- it looks like self.maz_ceiling can be either a pandas object or plain int. If the latter, the line below fails because the .astype method/function is not available. 

https://github.com/camsys/activitysim/blob/generic-whale/activitysim/core/los.py#L313

stefancoe:
@jpn-- For two-zone systems, it seems like self.maz_ceiling is always going to be an int:
[...]

PR #643, opened 2023-02-07
recode mazs in maz-to-maz table

jpn--:
Addresses #640

### DIFF
--- a/activitysim/core/los.py
+++ b/activitysim/core/los.py
@@ -298,7 +298,9 @@ class Network_LOS(object):
                 df["OMAZ"] = recode_based_on_table(df["OMAZ"], "land_use")
                 df["DMAZ"] = recode_based_on_table(df["DMAZ"], "land_use")
 
-                df["i"] = df.OMAZ * self.maz_ceiling + df.DMAZ
+                df["i"] = df.OMAZ.astype(np.int32) * self.maz_ceiling.astype(
+                    np.int32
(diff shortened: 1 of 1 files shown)


jpn--:
@stefancoe We apparently don't have a CI test that sufficiently exercises this code to trigger the bug, and similarly we don't have a test that would verify this change will fix it, but I think it might be as simple as this.  Can you test this update and confirm it fixes your code?  If so, we'll merge this, and then maybe flag this to be something to be prioritized for actual unit testing when we're ready for that in a few months.

stefancoe:
@jpn-- it fails on line 302 because the calculation on [line 301](https://github.com/camsys/activitysim/blob/bugfix-640/activitysim/core/los.py#L301) in los.py results in a int16 overflow, but I got it to work by changing it to this:

`df["i"] = df.OMAZ.astype(np.int32) * self.maz_ceiling.astype(np.int32) + df.DMAZ.astype(np.int32)`

stefancoe:
@jpn-- it looks like when running with legacy config settings, self.maz_ceiling is an int instead of a numpy.int32 and `self.maz_ceiling.astype(np.int32)` fails. This seems to work for all cases:

`df["i"] = df.OMAZ.astype(np.int32) * self.maz_ceiling + df.DMAZ.astype(np.int32)`

jpn--:
Superceded by #654

Issue #645, opened 2023-02-07
New Data Pipeline File Format


jpn--:
CS will change ActivitySim’s default data storage option for tabular data from HDF5 to Parquet. CS will benchmark associated ActivitySim performance improvement by documenting the resulting changes in file sizes and read/write times, using the MTC full scale example model.  Consultant will ensure that users will continue to be able to use HDF5 files if they want via a switch in the top-level settings, and will also enable HDF data compression options.

stefancoe:
The release of Pandas 2.0 includes Apache Arrow as an optional backend and it will automatically be used when reading from parquet files. This could result in performance improvements, including reduced RAM usage as string columns will use the Apache Arrow string type instead of the Python object type (If I am understanding things correctly). @jpn-- , @i-am-sijia any thoughts on these impacts? Perhaps we can touch on this at an upcoming Consortium meeting if it seems like improvements could be non-trivial. 

https://arrow.apache.org/blog/2019/02/05/python-string-memory-0.12/
https://datapythonista.me/blog/pandas-20-and-the-arrow-revolution-part-i

PR #649, opened 2023-02-07
parquet pipeline

jpn--:
Initial implementation for #645

### DIFF
--- a/activitysim/core/pipeline.py
+++ b/activitysim/core/pipeline.py
@@ -4,6 +4,7 @@ import datetime as dt
 import logging
 import os
 from builtins import map, next, object
+from pathlib import Path
 
 import pandas as pd
 from orca import orca
@@ -37,7 +38,7 @@ class Pipeline(object):
     def __init__(self):
         self.init_state()
 
-    def init_state(self):
+    def init_state(self, pipeline_file_format="parquet"):
 
         # most recent checkpoint
         self.last_checkpoint = {}
@@ -72,7 +73,7 @@ def is_open():
 def is_readonly():
     if is_open():
         store = get_pipeline_store()
-        if store and store._mode == "r":
+        if store and not isinstance(store, Path) and store._mode == "r":
             return True
     return False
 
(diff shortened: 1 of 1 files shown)


activitysim/core/pipeline.py
```diff
@@ -99,7 +100,11 @@ def close_open_files():
 
 def open_pipeline_store(overwrite=False, mode="a"):
     """
-    Open the pipeline checkpoint store
+    Open the pipeline checkpoint store.
+
+    If the pipeline_file_name setting ends in ".h5", then the pandas
+    HDFStore file format is used, otherwise pipeline files are stored
+    as parquet files organized in regular file system directories.
```
dhensle, line 107: I assume documentation on this setting will to be addressed in the other Pydantic task?

activitysim/core/pipeline.py
```diff
@@ -193,7 +216,11 @@ def write_df(df, table_name, checkpoint_name=None):
     We store multiple versions of all simulation tables, for every checkpoint in which they change,
     so we need to know both the table_name and the checkpoint_name to label the saved table
 
-    The only exception is the checkpoints dataframe, which just has a table_name
+    The only exception is the checkpoints dataframe, which just has a table_name,
+    although when using the parquet storage format this file is stored as "None.parquet"
+    to maintain a simple consistent file directory structure.
+
+    If the
```
dhensle, line 223: Looks like unfinished thought here...

activitysim/core/pipeline.py
```diff
@@ -209,10 +236,28 @@ def write_df(df, table_name, checkpoint_name=None):
...
     store = get_pipeline_store()
-
-    store[pipeline_table_key(table_name, checkpoint_name)] = df
-
-    store.flush()
+    if isinstance(store, Path):
+        store.joinpath(table_name).mkdir(parents=True, exist_ok=True)
+        df.to_parquet(store.joinpath(table_name, f"{checkpoint_name}.parquet"))
+    else:
+        complib = config.setting("pipeline_complib", None)
```
dhensle, line 243: Another setting to not get lost in the Pydantic task.

dhensle (commented):
Generally looks good and seems to work as designed for me.  Wondering if the default pipeline_file_name setting should change in config.py to remove the ".h5". (Had to do some digging here for the actual setting to get this feature turned on!).  Note that I think changing the default setting will also use this parquet functionality in the pipeline CI testing, which I think is desirable.  Based on the getting_started.ipynb notebooks, I think the .h5 version should be specified explicitly in the settings for the prototype_mtc example.

jpn--:
This PR was superseded by #654

Issue #644, opened 2023-02-07
Replace Orca


jpn--:
In order to support unit testing and other capabilities, the CS will eliminate the current ActivitySim dependency on the external Orca library.  CS will refactor Activity core functions and model steps to reflect a more typical Python style, passing in required variables as inputs in order to eliminate the reliance on orca to manage the program’s state.

PR #654, opened 2023-02-16
Replace ORCA with non-global state

jpn--:
Addresses #644

This PR has changes to the code in nearly every file, and removes the dependency on orca.  Several more advanced and complex features of orca have been entirely removed (e.g. extracting only certain columns in table merges), as they were never actually used in ActivitySim.

- [`core.workflow` documentation preview](https://camsys.github.io/activitysim/generic-whale/dev-guide/core-workflow.html)


jpn--:
note on [bc1a8be](https://github.com/ActivitySim/activitysim/pull/654/commits/bc1a8bed10d50ce7c4d057a96bf51199750c286e): the cdap estimation results for testing have changed slightly because the function assign_cdap_rank is used inside the estimation script.  It has no State object there so a simple stable random number generator is used to assign cdap ranks -- these random numbers are reproducible but different from the "reproducible random" numbers generated for a fully created State object.  So the results change a tiny bit.

activitysim/cli/run.py
```diff
@@ -154,68 +168,78 @@ def inject_arg(name, value, cache=False):
...
-            config_paths.insert(0, "configs_mp")
-            inject_arg("configs_dir", config_paths)
-
-        config.override_setting("multiprocess", True)
-        if args.multiprocess > 0:
-            config.override_setting("num_processes", args.multiprocess)
+    try:
+        state.load_settings()
+    except Exception as err:
+        logger.exception("erroro")
```
dhensle: Suggest a more detailed error message here.

activitysim/examples/prototype_mtc/configs_mp/logging.yaml
```diff
@@ -40,15 +40,18 @@ logging:
 
     logfile:
       class: logging.FileHandler
-      filename: !!python/object/apply:activitysim.core.config.log_file_path ['activitysim.log']
+      filename:
+        get_log_file_path: 'activitysim.log'
```
dhensle, line 44: From what I can tell, if we had a model setup with filename: activitysim.log instead of this change, then our model run with this code will no longer output a log file! This seems like a breaking change that needs to be either fixed or made very clear in the release notes.

activitysim/examples/prototype_mtc_extended/test/test_mtc_extended.py
dhensle, line 1: The prototype_mtc_extended example produces the following error:
147 od pairs not in skim including [[0 0 0 0 0]]->[[0 1 2 3 4]]

To reproduce, run with the following args:
        "args": [
            "-c", "activitysim/examples/prototype_mtc_extended/configs",
            "-c", "activitysim/examples/prototype_mtc/configs",
            "-d", "activitysim/examples/prototype_mtc/data",
            "-o", "activitysim/examples/prototype_mtc_extended/output",
        ]
Running with these args works.  Do not understand why...
            "args": [
                "-c", "activitysim/examples/prototype_mtc_extended/test/configs",
                "-c", "activitysim/examples/prototype_mtc_extended/configs",
                "-c", "activitysim/examples/prototype_mtc/configs",
                "-d", "activitysim/examples/prototype_mtc/data",
                "-o", "activitysim/examples/prototype_mtc_extended/test/output",
            ]

activitysim/abm/models/disaggregate_accessibility.py
```diff
@@ -740,24 +758,26 @@ def compute_disaggregate_accessibility(network_los, chunk_size, trace_hh_id):
...
-    for name, func in inject._DECORATED_TABLES.items():
-        logger.debug("reinject decorated table %s" % name)
-        orca.add_table(name, func)
+    for trace in [x for x in state.tracing.traceable_tables if "proto_" not in x]:
+        state.tracing.deregister_traceable_table(trace)
+
+    # # need to clear any premature tables that were added during the previous run
+    for name in list(state.existing_table_status):
+        if name not in tables_prior:
+            state.drop_table(name)
```
dhensle, line 766: I think we can get rid of the above lines 757-761 since we are deleting them here anyways.  Also, it looks like tables_prior includes the proto_population.  I don't think we need to keep these tables in the pipeline soaking up resources.  For that matter, I don't think we need to keep any tables in the pipeline except the "logsums" that are added below.
Also, the orca comments below can be cleaned up.

activitysim/abm/tables/shadow_pricing.py
```diff
@@ -1231,16 +1232,19 @@ def load_shadow_price_calculator(model_settings):
...
+) -> None:
+    _add_size_tables(state, disaggregate_suffixes)
 
 
 # then define _add_size_tables as a second method which also offers an optional
 # default argument to not scale sizes.  This is used only in disaggregate
 # accessibility (for now) and is not called via orca.  We need to do this to
 # avoid having to create a new orca variable for the scale argument.
-def _add_size_tables(disaggregate_suffixes, scale=True):
+def _add_size_tables(state, disaggregate_suffixes, scale=True) -> None:
```
dhensle: Do we still need this weird second method now that orca is gone?

dhensle (changes requested):
I figure this thing is big enough that I should do this in rounds so that things can be addressed while waiting for my continued review.

General question on ActivitySim style: should we be requiring type hints in our function definitions now? Cause it's a little bit all over the place right now.

jpn--:
> I figure this thing is big enough that I should do this in rounds so that things can be addressed while waiting for my continued review.

This is the way.

> General question on ActivitySim style: should we be requiring type hints in our function definitions now? Cause it's a little bit all over the place right now.

I have opened a new issue https://github.com/ActivitySim/activitysim/issues/682 for discussion on this and put my thoughts there.

activitysim/examples/prototype_mtc/configs_mp/logging.yaml
jpn--: I do not believe you will get a lack of log file as you describe.  I wrote a test to confirm this.  https://github.com/camsys/activitysim/blob/generic-whale/activitysim/core/test/test_logging.py

I also added documentation around recent changes for configuring logging.
https://github.com/camsys/activitysim/blob/939f323d7d094fc512402041d34ebc90bc682f61/docs/dev-guide/logging.md?plain=1#L23

activitysim/abm/tables/shadow_pricing.py
jpn--: Nope!  All gone

activitysim/cli/run.py
jpn--: added more detail.  also made it so errors here actually stop model execution (very early) so they can be fixed instead of bulldozed over

activitysim/abm/models/disaggregate_accessibility.py
jpn--: I dropped the dropping of two named tables as you suggest, and left the cleaning-up to the automatic filter in tables_prior.

I feel like cleaning up / dropping the proto_population tables would best be done as an explicit separate step, kind of like how `initialize_proto_population` is its own thing.  That will make it clearer what to do should anyone want to do something with the proto_pop after running `compute_disaggregate_accessibility`.

activitysim/examples/prototype_mtc_extended/test/test_mtc_extended.py
jpn--: Looks like this is happening because multiple tests are using the same output directory, and there is unexpected mixing of recoded (zero-based) and unrecoded (one-based) zone data.  For example, you think you are running with sharrow off and no recording of zone numbers, but the land use data is loaded out of a checkpoint file that did have zone recoding applied and has a "0" zone that isn't in the skims you have.

I modified the test file to use different output directories for sharrow-on and sharrow-off (and for shadow pricing for good measure) which should cure this condition in this particular case.  A more general solution to stop ActivitySim from being able to read data out of a mismatched checkpoint file is much more complicated and left to future work.

activitysim/core/chunk.py
```diff
@@ -649,41 +654,39 @@ def get_hwm_bytes(self):
...
     # check local hwm for all ledgers
     with ledger_lock:
         for c in CHUNK_LEDGERS:
             c.check_local_hwm(hwm_trace_label, rss, uss, total_bytes=None)
 
 
-def log_df(trace_label, table_name, df):
-
-    if chunk_training_mode() in (MODE_PRODUCTION, MODE_CHUNKLESS):
+def log_df(state: workflow.State, trace_label: str, table_name: str, df: pd.DataFrame):
```
i-am-sijia: I believe this method `log_df` line 680-701 can be deleted, given you have moved it under ChunkSizer and ChunkLedger, and those are being referenced in other modules.

activitysim/core/chunk.py
```diff
@@ -162,28 +156,17 @@ def chunk_logging():
...
 
 
-def default_initial_rows_per_chunk():
+def min_available_chunk_ratio(state: workflow.State):
     return SETTINGS.setdefault(
-        "default_initial_rows_per_chunk",
-        config.setting(
-            "default_initial_rows_per_chunk", DEFAULT_INITIAL_ROWS_PER_CHUNK
-        ),
+        "min_available_chunk_ratio", state.settings.min_available_chunk_ratio
```
i-am-sijia: `min_available_chunk_ratio` is not defined in the Settings class. I think the solution is to add it in "core/configuration/top.py" Settings class? This error showed up when I tried chunking the prototype_mtc example
```python
[00:07.02] ERROR: ===== ERROR IN initialize_landuse =====
[00:07.02] ERROR: 'Settings' object has no attribute 'min_available_chunk_ratio'
Traceback (most recent call last):
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\workflow\steps.py", line 46, in wrapper
    return func(*args, **kwargs)
  File "c:\users\wangs1\workspace\activitysim\activitysim\abm\models\initialize.py", line 116, in initialize_landuse
    with chunk.chunk_log(state, trace_label, base=True) as chunk_sizer:
  File "C:\Users\wangs1\workspace\ASIM-DEV-py10\lib\contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 1109, in chunk_log
    chunk_sizer = ChunkSizer(
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 740, in __init__
    min_chunk_ratio = min_available_chunk_ratio(self.state)
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 161, in min_available_chunk_ratio
    "min_available_chunk_ratio", state.settings.min_available_chunk_ratio
AttributeError: 'Settings' object has no attribute 'min_available_chunk_ratio'
```

activitysim/core/chunk.py
```diff
@@ -649,41 +654,39 @@ def get_hwm_bytes(self):
...
         # mem.trace_memory_info(hwm_trace_label, trace_ticks=trace_ticks)
         return
 
-    rss, uss = mem.trace_memory_info(hwm_trace_label)
+    rss, uss = mem.trace_memory_info(hwm_trace_label, state=state)
 
     # check local hwm for all ledgers
     with ledger_lock:
         for c in CHUNK_LEDGERS:
             c.check_local_hwm(hwm_trace_label, rss, uss, total_bytes=None)
```
i-am-sijia:
```suggestion
            c.check_local_hwm(state, hwm_trace_label, rss, uss, total_bytes=None)
```

activitysim/core/chunk.py
```diff
@@ -1039,16 +1046,79 @@ def ledger(self):
...
+            # trace_ticks = 0 if force else mem.MEM_TRACE_TICK_LEN
+            # mem.trace_memory_info(hwm_trace_label, trace_ticks=trace_ticks)
+            return
+
+        rss, uss = mem.trace_memory_info(hwm_trace_label, state=self.state)
+
+        # check local hwm for all ledgers
+        with ledger_lock:
+            for c in CHUNK_LEDGERS:
+                c.check_local_hwm(hwm_trace_label, rss, uss, total_bytes=None)
```
i-am-sijia: missing state argument?
```suggestion
                c.check_local_hwm(self.state, hwm_trace_label, rss, uss, total_bytes=None)
```

activitysim/core/chunk.py
```diff
@@ -1039,16 +1046,79 @@ def ledger(self):
...
+
+        # registers this df and recalc total_bytes
+        cur_chunker.log_df(table_name, df)
+
+        total_bytes = sum([c.total_bytes for c in CHUNK_LEDGERS])
+
+        # check local hwm for all ledgers
+        with ledger_lock:
+            for c in CHUNK_LEDGERS:
+                c.check_local_hwm(hwm_trace_label, rss, uss, total_bytes)
```
i-am-sijia:
```suggestion
                c.check_local_hwm(self.state, hwm_trace_label, rss, uss, total_bytes)
```

activitysim/core/chunk.py
i-am-sijia: seems missing the state argument in all lines that calls check_local_hwm()?

activitysim/core/chunk.py
```diff
@@ -710,15 +713,24 @@ def run(self):
             log_rss(self.trace_label)
```
i-am-sijia: This will error out since it's missing the state argument. Same for line 711 above.
```python
[00:11.67] INFO: #run_model running step initialize_landuse
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\wangs1\workspace\ASIM-DEV-py10\lib\threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 711, in run
    log_rss(self.trace_label)
TypeError: log_rss() missing 1 required positional argument: 'trace_label'
[00:11.90] INFO: Reading CSV file c:\users\wangs1\workspace\activitysim\activitysim\examples\prototype_mtc\data\land_use.csv
[00:11.99] INFO: loaded land_use (25, 24)
[00:12.31] ERROR: mem_monitor for initialize_landuse.chunk_log died!
```

activitysim/core/chunk.py
```diff
@@ -1039,16 +1046,79 @@ def ledger(self):
...
+
+        op = "del" if df is None else "add"
+        hwm_trace_label = f"{trace_label}.{op}.{table_name}"
+
+        rss, uss = mem.trace_memory_info(hwm_trace_label, state=self.state)
+
+        cur_chunker = CHUNK_LEDGERS[-1]
+
+        # registers this df and recalc total_bytes
+        cur_chunker.log_df(table_name, df)
```
i-am-sijia: This will error out since it is missing the state argument 
```python
[00:12.31] ERROR: mem_monitor for initialize_landuse.chunk_log died!
[00:12.31] ERROR: ===== ERROR IN initialize_landuse =====
[00:12.32] ERROR: bug
Traceback (most recent call last):
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 1026, in ledger
    yield
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 1139, in chunk_log
    yield chunk_sizer
  File "c:\users\wangs1\workspace\activitysim\activitysim\abm\models\initialize.py", line 121, in initialize_landuse
    annotate_tables(state, model_settings, trace_label, chunk_sizer)
  File "c:\users\wangs1\workspace\activitysim\activitysim\abm\models\initialize.py", line 67, in annotate_tables
    chunk_sizer.log_df(trace_label, tablename, df)
  File "c:\users\wangs1\workspace\activitysim\activitysim\core\chunk.py", line 1085, in log_df
    cur_chunker.log_df(table_name, df)
TypeError: ChunkLedger.log_df() missing 1 required positional argument: 'df'
```

activitysim/examples/prototype_mtc/configs_mp/logging.yaml
i-am-sijia: Adding to @dhensle 's comment. When I run example MTC without multiprocessing, I don't see the activitysim.log file being generated in the output folder, but being generated in whichever folder I'm calling test_mtc.py from.

i-am-sijia (changes requested):

i-am-sijia:
Somehow the blurb of my review summary did not show up, so retyping it here:

I have reviewed:

- cdap with joint tour utility
- joint tour frequency and composition
- non-mandatory tour frequency
- parking location
- chunk.py
- state.py
- and a few other .py files in the core folder

One main comment is that chunking does not seem to be working with the new code, and it seems that none of the current test examples includes chunking. I tried running the prototype_mtc with chunking turned on and it errored out at multiple places, as noted in my review comments. I tried to fix them as I go, but did not fully debug it through nor get chunking running with prototype_mtc. @jpn-- , have you tested chunking with any of the example models? Am I missing anything?

activitysim/core/cleaning.py
```diff
@@ -32,24 +34,35 @@ def recode_to_zero_based(values, mapping):
     return result
 
 
-def should_recode_based_on_table(tablename):
+def should_recode_based_on_table(state: workflow.State, tablename):
     try:
-        base_df = inject.get_table(tablename).to_frame()
+        base_df = state.get_dataframe(tablename)
     except (KeyError, RuntimeError):
         # the basis table is missing, do not
```
dhensle, line 41: This try-except block is actually capturing an error that I think should cause a crash.  If a taz.csv file is not supplied, a runtime error is raised but this block captures it and the run continues.  Its unclear exactly why a taz.csv file is even necessary -- we should be able to get a unique list of taz's from the landuse table. 

To reproduce, run the PSRC example by re-naming taz.csv input table to something else, e.g. taz_.csv, and remove taz.csv from the settings.yaml input tables. (Not all models have an input taz.csv file, which is how I found the error, but this is the easiest way to replicate.)

I think there are two things to fix here:
1. taz.csv looks to be optional unless running with sharrow.  Is that True? If so, why? If not supplied, can taz.csv default to grab a unique list from the landuse taz column?
2. If an error message is thrown to the log, ActivitySim should crash.  If the file is optional, then a warning should be sent instead. It is very confusing for the user to have an error message thrown that doesn't kill the run.

activitysim/core/configuration/filesystem.py
```diff
@@ -0,0 +1,828 @@
...
+
+    def read_model_coefficients(self, model_settings=None, file_name=None):
+        from activitysim.core import simulate
+
+        return simulate.read_model_coefficients(
+            self, model_settings=model_settings, file_name=file_name
+        )
+
+    def get_segment_coefficients(self, model_settings, segment_name):
+        from activitysim.core import simulate
```
dhensle, line 826: Can we move all these simulate imports to the top of the file?

activitysim/core/configuration/top.py
```diff
@@ -457,3 +547,98 @@ class Settings(PydanticBase):
...
+    * "raise"
+        Attempts to re-run a step that has already been run and
+        checkpointed will raise a `RuntimeError`, halting model execution.
+        This is the default if no value is given.
+    * "warn"
+        Attempts to re-run a step that has already been run and
+        checkpointed will be trigger a warning message and that particular step
+        will not be (re)executed, but overall model execution will be allowed to
+        continue.
+    * "allow"
```
dhensle, line 633: Are there any other unforeseen circumstances to allowing this besides checkpointing?  It also doesn't seem to actually work.  I get the following error when adding auto ownership again after, e.g. cdap. `UserWarning: aborting attempt to re-run step 'auto_ownership_simulate' more than once` and there's no evidence for auto ownership being run again.

I tried auto ownership because it seems like one of the few use cases for this functionality: pre-seeding auto ownership before location choice and then running again after workplace/school determined.

activitysim/core/contrast/continuous.py
dhensle, line 1: Is there any documentation on what all this contrast code is about besides [this little blurb](https://github.com/camsys/activitysim/blob/generic-whale/docs/dev-guide/workflows.md#workflows-for-performance-monitoring)? (The [sharrow-contrast link](https://github.com/camsys/activitysim/blob/generic-whale/docs/dev-guide/workflows.md#workflows-for-performance-monitoring:~:text=found%20in%20the-,sharrow%2Dcontrast,-workflow%20subdirectory.%20Each) in there is outdated too.) Either way I think additional documentation here about what this is all for and how to configure would be helpful.

activitysim/core/exceptions.py
dhensle, line 1: I like this! Creating ActivitySim specific exceptions with custom error messages is a good idea! Woud it be useful to add a ZeroProbabilityError with some better documentation and move some of that report_bad_choices() stuff in logit.py over here?  Having a single place to document common ActivitySim errors like this seems very useful.

dhensle (changes requested):
Another round.

activitysim/examples/prototype_mtc/configs_mp/logging.yaml
jpn--: Looks like the problem @i-am-sijia identified here is a config issue not a code issue; the plain (non-mp) config gave the log file as 

```
filename: activitysim.log
```

Which just dumps the logs to CWD, as noted.  The preferred pattern is to always use

```
filename:
  get_log_file_path: 'activitysim.log'
```

which puts the file where we expect, whether single or multi-processing.  I changed the logging.yaml for the single process to do this.

activitysim/core/cleaning.py
jpn--: Yes, the land_use_taz content is optional unless you are recoding zone numbers, and the recoding was introduced with Sharrow, although in theory you can apply the recoding and not otherwise use sharrow (but I'm not sure why you would want to).

I've modified the land_use_taz table to automatically infer the TAZs from the land use table instead of logging an error that is then handled, although there will need to be some consistency with the skims -- the TAZs must match, and I think that consistency is currently trust-based and not otherwise checked.

activitysim/core/configuration/filesystem.py
jpn--: Moving all the imports to the top would be great!  And it's definitely on the to-do list. Unfortunately this code has gotten wound around itself in my effort to make these changes substantially backward compatible and keep everything working through the transition, such that putting the imports at the top level triggers a circular import that crashes ActivitySim.

activitysim/core/configuration/top.py
jpn--: @dhensle  Are you getting the "aborting attempt..." log message when `duplicate_step_execution` is set to "allow"?  You should only see it for when you have set to "warn", and in that case it is the correct behavior that the model should not be run again, but also it should not stop the overall execution of the system.  Only for "allow" should it try to run again, and you should get no "aborting attempt" message.

activitysim/examples/prototype_mtc_extended/test/test_mtc_extended.py
dhensle: Looks to be resolved. Thanks

activitysim/core/cleaning.py
dhensle: Suggest adding an assert statement when inferring TAZs from the land use table (and in general?) that the number of TAZs matches the shape of the skims.

activitysim/core/configuration/filesystem.py
dhensle: Ok, sounds like an issue should be created for this since it's not going to be addressed here? It would be good if the issue could provide a little bit more detail about the problem area(s) and temporary fixes for developers to try if they run into the issue (switching the import order of some packages, etc.)

activitysim/core/configuration/top.py
dhensle: Ah, I was misunderstanding what "warn" means.  The text description here is confusing: "attempts to re-run a step.... and that particular step will not be (re)executed."  Do we need a "warn" option?  I am not sure I see the utility of this -- either we want to allow it or not.  Either way, I think the description here needs to be updated. 

In a similar vein, the data dictionary only shows the first instance a variable shows up, but if that is modified in downstream models, there is no record of that in the output data_dict.txt file.  I think we should provide duplicates, especially if we are allowing models to be re-run now.

activitysim/core/workflow/tracing.py
```diff
@@ -0,0 +1,741 @@
...
+            ["households", "proto_households"]
+        )
+
+        assert len(persons_table_name) == 1 and len(persons_table_name) == 1
+        persons_table_name, households_table_name = (
+            persons_table_name.pop(),
+            households_table_name.pop(),
+        )
+
+        if (
```
dhensle, line 523: This needs to be updated to trace the proto_population in disaggregate accessibility.  Trying to trace right now crashes.  You can just copy over the code from [here](https://github.com/SANDAG/activitysim/blob/BayDAG_estimation/activitysim/core/tracing.py#L779) (lines 779 to 805) and it seems to work fine.

activitysim/core/workflow/util.py
```diff
@@ -0,0 +1,101 @@
...
+#     from .steps import _STEP_LIBRARY
+#     try:
+#         step_func = _STEP_LIBRARY[name]
+#     except KeyError:
+#         logger.error(f"Unknown step {name}, the known steps are:")
+#         for n in sorted(_STEP_LIBRARY.keys()):
+#             logger.error(f" - {n}")
+#         raise
+#     step_func(context)
+#     return context
```
dhensle: cleanup

dhensle (changes requested):
Seems like now might be a good time to sneak in a fix for the pandas performance warning issue: https://github.com/ActivitySim/activitysim/issues/685

activitysim/core/exceptions.py
jpn--: Agree this is a good plan.  I opened a new issue #692 to track this, but I don't want to hold this PR for peripherally related changes.

activitysim/core/contrast/continuous.py
jpn--: yeah, the `core.contrast` module is not documented and incomplete.  I am pulling it out of this PR, it's a great demo of the usefulness of having this PR but it's not actually needed here.  I'll open a new different PR for this code so we can finish the orca removal cleanly and soon, and we can deliver the visualization tools (which are beyond our scope) as an add on later.  Also, the outdated link is fixed.

activitysim/core/configuration/top.py
jpn--: on reflection, I agree the "warn" is confusing and unnecessary, so I'm dropping it as an option here, leaving only "error" (renamed from "raise" to maybe be clearer) and "allow".

I'm going to defer the data dictionary issue for future work, as it overlaps heavily with the data model stuff, and it's not limited to just re-running the same component, there can be instance were a different component touches the same data column, e.g. the atwork subtour mode and destination models touch things but don't get listed in that file either.

Issue #593, opened 2022-08-29
Re-Structure of Trip Scheduling


dhensle:
The current implementation of probabilistic trip scheduling results in about 25% of stops failing to produce valid schedules.  Trips fail to schedule when a selection for an earlier trip in a half-tour makes scheduling of a subsequent trip impossible (or very low probability) based on the probability distribution.

The current [trip scheduling probability distributions](https://github.com/ActivitySim/activitysim/blob/develop/activitysim/examples/prototype_mtc/configs/trip_scheduling_probs.csv) are segmented by purpose, half-tour direction, tour time period, and trip number.  The selected probabilities are independent of what the previous trip departure time was and can lead to the failed trips. The current implementation is consistent with TM1's version of CT-RAMP.

Re-structuring the trip scheduling probabilities file to replace the tour time period segment with the previous trip's time period would ensure that a trip does not get scheduled before it's previous trip. This enhancement was made in the SANDAG version of CT-RAMP and an example probabilities table for that implementation can be seen [here](https://github.com/SANDAG/ABM/blob/ABM3_develop/uec/StopDepartArriveProportions.csv).

The advantages of this new approach would be less failed trip schedules.  Trip scheduling currently iterates up to the [maximum iterations](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/prototype_mtc/configs/trip_scheduling.yaml#L11) to try to get a valid tour pattern, so less failed trips would also mean less iteration and therefore decreased run-times.

(A [setting](https://github.com/ActivitySim/activitysim/blob/main/activitysim/examples/prototype_mtc/configs/trip_scheduling.yaml#L14) in the current implementation will deal with failed trips by either throwing them out or setting them to the previous trip departure time in the outbound direction or next trip departure time in the inbound direction.)

lmz:
From conversation with @DavidOry, it is my understanding that there are multiple approaches:
1) CTRAMP2 method (which is much more elaborate, includes accessibilities -- this is too significant a difference and is unlikely to be considered) 
2) Atlanta ActivitySim model approach
3) SANDAG approach referenced above

It would be helpful to have a description of 2 vs 3 and pros/cons described here.

DavidOry:
[...]

Issue #587, opened 2022-08-18
Extension of Vehicle Type Model to Include AVs


dhensle:
RSG is extending ActivitySim to consider autonomous vehicles (AVs) for consortium partners MWCOG and SANDAG.  One implementation option is to treat AV's as an extension of body type in the vehicle type model.  e.g. car would be split into car and car_av.  This new body type would just have to be included in the [vehicle type data file](https://github.com/ActivitySim/activitysim/blob/master/activitysim/examples/prototype_mtc_extended/configs/vehicle_type_data.csv) and would require minimal code changes.

When including a new body type in the vehicle type data file, there is [a check](https://github.com/ActivitySim/activitysim/blob/master/activitysim/abm/models/vehicle_type_choice.py#L217) to make sure the vehicle type data exists for every [alternative](https://github.com/ActivitySim/activitysim/blob/master/activitysim/examples/prototype_mtc_extended/configs/vehicle_type_choice.yaml#L37).  This was a design decision to make sure the user assembling the vehicle type data file didn't miss any alternatives, especially when extending to future years.  However, this results in a set of alternatives in the base-year that is more than twice as big as it needs to be, because there are alternatives for vehicles that did not exist in previous years (Hybrids, PEVs, HEVs). Though these alternatives are unavailable they still more than double the amount of RAM required for running the model.

Extending the body type to include AVs would again double the number of alternatives required, resulting in 4x the amount of RAM required for the model in the base-year, and more than double the necessary amount of RAM for future year files since there are many combinations of AV and fuel type that are not necessary.

We propose either putting a switch in the model yaml to distinguish between two options:
1. Strictly enforce that all alternatives have a matching row in the vehicle type data (which is how it currently works)
2. Just remove alternatives from the model that are not supplied with vehicle type data

Or, simply remove the assert statement from the code.

bettinardi:
I'm curious if this issue exists for both implementations of the vehicle model.  I forget what they are called now (we should ensure we have good names), but during construction I believe the two implementations that moved forward were option 4 and 2.  Maybe we call them detailed and simplified or something.  Anyway, I hope it is the intent to ensure that AV specification works for both options, and I would like to verify whether this issues exists for both options or just option 4 (detailed).

[...]

PR #657, opened 2023-03-09
BayDAG Contributions

dhensle:
This pull request is a joint effort between WSP & RSP working on the MTC and SANDAG models respectively.

Changes include:
* Joint tour extension to CDAP
* New joint tour frequency and composition model 
* "relative" trip scheduling (#593)
* Option to remove constraint requiring data for every single alternative in vehicle type model (#587)
* Fixing overflow error in non-mandatory tour frequency model alternatives (int8 -> int16)
* Annotate functionality in trip mode choice
* Unit testing infrastructure used to test conversion of CT-RAMP UECs to ActivitySim configs


i-am-sijia:
Looks good to me.

PR #660, opened 2023-03-23
Trip scheduling logic

jpn--:
The BayDAG pull request included a change in trip scheduling that alters the results of the model for certain corner cases.  The change is not detectable in the current CI test suite, which only tests a very limited number of households and does not find many corner-case errors.  (The upcoming refactoring includes infrastructure to rigorously test much larger subsets of models, which is how this was discovered.  For the MTC test model, the change impacts the departure time of approximately 0.7% of all simulated trips.)

This PR rolls back the model component to use the original logic by default, and allows users to upgrade to the revised logic by setting "logic_version: 2" in the model settings.  If no logic level is set explicitly, a warning is emitted that "logic_version: 1" is implied, and that a future version of ActivitySim may default to logic version 2.

### DIFF
--- a/activitysim/abm/models/trip_scheduling.py
+++ b/activitysim/abm/models/trip_scheduling.py
@@ -327,7 +327,12 @@ def schedule_trips_in_leg(
             # choices are relative to the previous departure time
             choices = nth_trips.earliest + choices
             # need to update the departure time based on the choice
-            update_tour_earliest(trips, choices, _logic_version(model_settings))
+            logic_version = _logic_version(model_settings)
+            if logic_version == 1:
+                raise ValueError(
+                    "cannot use logic version 1 with 'relative' scheduling mode"
+                )
+            update_tour_earliest(trips, choices, logic_version)
 
         # adjust allowed depart range of next trip
         has_next_trip = nth_trips.next_trip_id != NO_TRIP_ID
--- a/activitysim/abm/models/trip_scheduling.py
+++ b/activitysim/abm/models/trip_scheduling.py
@@ -1,6 +1,7 @@
 # ActivitySim
 # See full license in LICENSE.txt.
 import logging
+import warnings
 from builtins import range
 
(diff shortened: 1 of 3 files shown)


activitysim/abm/models/trip_scheduling.py
```diff
@@ -145,11 +170,18 @@ def update_tour_earliest(trips, outbound_choices):
...
+            tmp_trips["outbound"], tmp_trips["max_outbound_departure"]
+        )
+    elif logic_version > 1:
+        tmp_trips["earliest"] = np.where(
+            ~tmp_trips["outbound"] & ~tmp_trips["max_outbound_departure"].isna(),
+            tmp_trips["max_outbound_departure"],
+            tmp_trips["earliest"],
+        )
+    else:
+        raise ValueError(f"bad logic_version: {logic_version}")
```
dhensle, line 184: I think we need a check somewhere in here that will require logic_version == 2 if the scheduling mode is "relative", otherwise that mode will fail. Should mention this requirement in the documentation as well.

dhensle (changes requested):
Code looks fine to me minus my one comment. A lot of changes to keep some "broken" functionality, but I understand the desire to keep the previous large-scale versions unchanged.

activitysim/abm/models/trip_scheduling.py
jpn--: I've added that check to the code, and a note in the documentation.

jpn--:
Thanks @dhensle.  I agree there's a lot of dancing around here to preserve a "broken" model, but this is the price we need to pay to keep everyone on one platform.

activitysim/abm/models/trip_scheduling.py
dhensle: Looks good, thanks @jpn-- .  (And thanks for fixing my spelling errors in the documentation too!)

PR #661, opened 2023-03-25
Moved 'tot_tours' from nm tour frequency script to alternatives file

### LINKED ISSUE #656: tot_tours Hidden in Non-Mandatory Tour Frequency Code
A variable named `tot_tours` is hard-coded in the non-mandatory tour frequency model: https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/non_mandatory_tour_frequency.py#L159.  
This should really be exposed to the user and listed in the alternatives file instead.

JoeJimFlood:
To resolve #656


activitysim/abm/models/non_mandatory_tour_frequency.py
```diff
@@ -151,12 +151,9 @@ def non_mandatory_tour_frequency(persons, persons_merged, chunk_size, trace_hh_i
 
     model_settings = config.read_model_settings(model_settings_file_name)
 
-    # FIXME kind of tacky both that we know to add this here and del it below
-    # 'tot_tours' is used in model_spec expressions
     alternatives = simulate.read_model_alts(
         "non_mandatory_tour_frequency_alternatives.csv", set_index=None
     )
-    alternatives["tot_tours"] = alternatives.sum(axis=1)
```
jpn--, line 208: I would leave this line in place, but rather than simply assigning to "tot_tours": (1) check if that column name already exists, (2) write this sum if does not exist, and emit a FutureWarning while doing so to inform the user that ActivitySim may not auto-populate the "tot_tours" value in the future, and (3) emit a warning (not a show-stopping error) if there is any consistency problems between an exogenously defined "tot_tours" and the expected values.  This will allow existing model implementations to survive the code change without editing config files.

dhensle:
I ran into an issue with this change awhile ago while doing some estimation work. [Canonical_ids](https://github.com/ActivitySim/activitysim/pull/581/files) (used to create reproducible tour/trip ids) calculates IDs based on purposes specified in the alts file.  If tot_tours is listed in the alt file, it is treated as another purpose. Thus, adding a column to input alts will change IDs and break estimation mode.

In other words, I believe this change should not pass the CI testing for estimation mode.  

I think we have a couple of options:
1. Explicitly specify the alts for canonical IDs in the example model(s) and keep the change as-is.
2. Update canonical IDs to avoid tot_tours being counted as a separate purpose.
3. Push this issue into data model dev task instead of phase 8. (could be done regardless of whether 1 or 2 is selected.)

jpn-- (approved):

jpn--:
I chose option 2, which was relatively easy.  A more comprehensive solution is deferred to whenever we address a "data model". 

> I ran into an issue with this change awhile ago while doing some estimation work. [Canonical_ids](https://github.com/ActivitySim/activitysim/pull/581/files) (used to create reproducible tour/trip ids) calculates IDs based on purposes specified in the alts file. If tot_tours is listed in the alt file, it is treated as another purpose. Thus, adding a column to input alts will change IDs and break estimation mode.
> 
> In other words, I believe this change should not pass the CI testing for estimation mode.
> 
> I think we have a couple of options:
> 
> 1. Explicitly specify the alts for canonical IDs in the example model(s) and keep the change as-is.
> 2. Update canonical IDs to avoid tot_tours being counted as a separate purpose.
> 3. Push this issue into data model dev task instead of phase 8. (could be done regardless of whether 1 or 2 is selected.)

PR #672, opened 2023-04-26
added stricter joining of annotated fields

ghost:
Added a check to prevent out of order pandas series from being concatenated improperly. If there is an index in the panda series then it is reindexed to match the final data frame, otherwise skip.

### DIFF
--- a/activitysim/core/test/test_util.py
+++ b/activitysim/core/test/test_util.py
@@ -7,7 +7,7 @@ import pandas as pd
 import pandas.testing as pdt
 import pytest
 
-from ..util import other_than, quick_loc_df, quick_loc_series, reindex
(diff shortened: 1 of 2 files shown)


dhensle:
Context here is that the preprocessor runs all the python expressions and then concatenates them. If they are all pandas with the same index, they'll get joined together no problem.  However, if for example it's an array that gets turned back into a pandas series but now has a different order, it'll be scrambled when you try to join them. 

In other words, this PR helps prevent users from an incorrect merge when writing complicated preprocessor expressions.

jpn--:
This PR seems like a great candidate for "show me in a unit test".  Can we do that in a timely manner here?

dhensle:
Unit test added and passing.  Ready for your review @jpn--.

jpn-- (approved):

PR #689, opened 2023-07-11
Bugfix for populationsim

ghost:
the "start_time" variable on line 1484 is unbound if sharrow is not enabled and there is no shared data buffer (as is the case for populationsim using the ActivitySim framework), it will result in an undefined start_time. 

A simple fix is to just get start_time once before the if-else.


jpn--:
This small fix has already been made in a different PR that has been merged to the develop branch.

PR #694, opened 2023-07-25
orca residual cleanup

jpn--:
remove leftover references to ORCA from docstrings and now-unused imports


jpn--:
This PR has no material code changes, it is just cleaning up residual references to ORCA in annotations, comments, and docstrings, so I am merging it to the develop branch now.

PR #751, opened 2023-09-26
Fix memory usage

jpn--:
- When caching sharrow flows, don't cache tree data implicitly embedded in the flow.
- Implement categorical dtype for some time period data.


jpn--:
@i-am-sijia You've tested this as part of memory usage work, and reported to the group that it seemed to work well.  Are you comfortable with all the code / ready to merge this PR?

i-am-sijia:
@jpn-- yes that's fine.

PR #759, opened 2023-11-23
Explicit chunking

jpn--:
This PR adds an explicit chunking option to the compute_accessibility component.


stefancoe (approved):
Code looks good. I tested with and without chunking using PSRC's implementation and am getting the same results. @jpn-- does the documentation need anything beyond what will get updated via documentation build?

jpn--:
>  does the documentation need anything beyond what will get updated via documentation build?

Yes, we do need to write a much better user's guide on how to use chunking.  Since this PR only includes explicit chunking for a single component, I've not taken that on yet (i.e. in this PR) but it does include documentation attached as docstrings to the "explicit_chunk" settings for the AccessibilitySettings class, and in the doc embedded in the chunk.py file.

Issue #762, opened 2023-12-12
Add option for output tables to be written as parquet files. 


stefancoe:
**Is your feature request related to a problem? Please describe.**
Current options include csv and h5. Parquet offers a better file based option when compared to csv, for both size and speed. PSRC's trip table as a csv is  2,100,000 KB compared to 459,700 KB as a parquet file.  Loading the csv file into a Pandas Dataframe on my laptop takes 20 seconds compared to 3 seconds as a parquet file.  Activitysim is already using parquet to store pipeline files.  

**Describe the solution you'd like**
Currently, there is a config setting called 'h5_store', that uses h5 when set to True and csv when set to False or not included. So csv is the default. I propose adding a setting called 'file_type' that would allow 3 options: 'csv', 'h5', or 'parquet'. Its default would also be 'csv'. The h5_store setting would remain and its current expected behavior would be unchanged. The behavior of these settings would work like so:

- When h5_store is set to True outputs are written out to h5. 
- When h5_store is set to False (default) and file_type is not specified, outputs are written as .csv
- When h5_store is set to False (default) and file_type is specified, outputs are written out to its setting: csv, parquet or h5. 
- file_type is validated against allowed values (csv, parquet, h5) using pydantic. Activitysim will crash with a useful error message almost immediately if this setting is included with a wrong value. 

**Describe alternatives you've considered**
Another option would be to add a boolean setting like use_parquet, but conflicts would arise if both settings were to set to True in a config file.  If this request is accepted and we go with file_type, it may make sense to deprecate the h5_store setting at some point, especially if even more file types are supported in the future. 

**Additional context**
I have made these changes on a fork and will issue a pull request.

PR #763, opened 2023-12-12
Option to write output tables as parquet files

stefancoe:
This PR adds the option to write output tables as parquet files, as described in this [issue/feature request](https://github.com/ActivitySim/activitysim/issues/762).


activitysim/core/configuration/top.py
```diff
@@ -119,6 +120,11 @@ class OutputTables(PydanticBase):
     h5_store: bool = False
     """Write tables into a single HDF5 store instead of individual CSVs."""
 
+    file_type: str = "csv"
```
jpn--: If we change this line to: `file_type: Literal["csv", "parquet", "h5"] = "csv"` then we don't need the seperate validator function, we get that check for free. and the code will be easier to read.

activitysim/core/configuration/top.py
```diff
@@ -144,6 +150,15 @@ class OutputTables(PydanticBase):
     applied to any output tables.
     """
 
+    @validator("file_type")
+    def method_is_valid(cls, method: str) -> str:
```
jpn--: drop this after changing type from `str` to `Literal`

activitysim/core/steps/output.py
```diff
@@ -226,8 +227,13 @@ def write_data_dictionary(state: workflow.State) -> None:
...
     """
-    Write pipeline tables as csv files (in output directory) as specified by output_tables list
-    in settings file.
+    Write pipeline tables as csv or parquet files (in output directory) as specified
+    by output_tables list in settings file. Output to parquet or a single h5 file is
+    also supported.
+
+    'h5_store' defaults to False, which means the output will be written out to csv.
+    'file_type' defaults to 'csv' but can also be used to specify 'parquet' or 'h5'.
+    When 'h5_store' is set to True, 'file_type' is ingored and the outputs are written to h5.
```
jpn--, line 236: The old setting should now be deprecated.  I opened an issue #791 to address that separate from this PR.

activitysim/core/steps/output.py
```diff
@@ -383,14 +400,18 @@ def map_func(x):
...
         else:
-            file_name = f"{prefix}{table_name}.csv"
+            file_name = f"{prefix}{table_name}.{file_type}"
             file_path = state.get_output_file_path(file_name)
 
             # include the index if it has a name or is a MultiIndex
-            csv.write_csv(dt, file_path)
+            if file_type == "csv":
+                csv.write_csv(dt, file_path)
+            else:
```
jpn--: if we make this `elif file_type == "parquet"` and add an `else: ValueError`, we'll have a clearer framework to add yet other formats in the future.

jpn-- (changes requested):
Looks good, just a few suggested small changes to make this code more future-developer-friendly.

PR #768, opened 2023-12-18
BayDAG Contribution #2: Enhanced Disaggregate Accessibility Merging

dhensle:
This pull request gives the user greater flexibility to join disaggregate accessibilities to the households / person tables. The original design for the model just joined the disaggregate accessibilities to the household table by a set of auto ownership and home zone.  This was insufficient for the purpose of joining all auto ownership levels created in the disaggregate accessibilities model so they could be applied to the alternatives in the auto ownership model.  

See the SANDAG ABM3 configs for an example implementation: [yaml](https://github.com/SANDAG/ABM/blob/ABM3_develop/src/asim/configs/resident/disaggregate_accessibility.yaml#L120) and [annotator](https://github.com/SANDAG/ABM/blob/ABM3_develop/src/asim/configs/resident/annotate_disaggregate_accessibility.csv).

Required for SANDAG ABM3 production? -- yes

### DIFF
--- a/activitysim/abm/models/disaggregate_accessibility.py
+++ b/activitysim/abm/models/disaggregate_accessibility.py
@@ -164,6 +164,11 @@ class DisaggregateAccessibilitySettings(PydanticReadable, extra="forbid"):
     """
     NEAREST_METHOD: str = "skims"
 
+    postprocess_proto_tables: list[DisaggregateAccessibilityAnnotateSettings] = []
+    """
+    List of preprocessor settings to apply to the proto-population tables after generation.
+    """
+
 
 def read_disaggregate_accessibility_yaml(
     state: workflow.State, file_name
@@ -847,7 +852,7 @@ def compute_disaggregate_accessibility(
         del df
 
     disagg_model_settings = read_disaggregate_accessibility_yaml(
-        "disaggregate_accessibility.yaml"
+        state, "disaggregate_accessibility.yaml"
     )
 
     # Run location choice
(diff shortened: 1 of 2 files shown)


AndrewTheTM:
Just commenting to say that these are necessary for Met Council's Phase 2 ActivitySim model. 

cc: @JonathanEhrlichMC, @dvfarmer

i-am-sijia (approved):

jpn--:
This non-trivial PR has no testing to confirm it works.

activitysim/abm/models/disaggregate_accessibility.py
```diff
@@ -164,6 +164,11 @@ class DisaggregateAccessibilitySettings(PydanticReadable, extra="forbid"):
     """
     NEAREST_METHOD: str = "skims"
 
+    postprocess_proto_tables: list[DisaggregateAccessibilityAnnotateSettings] = []
+    """
+    List of preprocessor settings to apply to the proto-population tables after generation.
+    """
+
```
jpn--, line 180: This PR actually adds two things to DisaggregateAccessibilitySettings, we also need to add something like this...

```
    KEEP_COLS: list[str] | None = None
    """
    Columns to keep in the proto-population tables after generation.

    If None, all columns that have names containing the word ""accessibility" are kept.
    """
```

activitysim/abm/tables/disaggregate_accessibility.py
```diff
@@ -172,14 +172,13 @@ def disaggregate_accessibility(state: workflow.State):
     accessibility_cols = [
         x for x in proto_accessibility_df.columns if "accessibility" in x
     ]
+    keep_cols = model_settings.get("KEEP_COLS", accessibility_cols)
```
jpn--: This should be something like:
```
    if model_settings.KEEP_COLS is None:
        keep_cols = [x for x in proto_accessibility_df.columns if "accessibility" in x]
    else:
        keep_cols = model_settings.KEEP_COLS
```

jpn-- (changes requested):

jpn-- (approved):

PR #769, opened 2023-12-18
BayDAG Contribution #3: CDAP estimation with joint component

dhensle:
As part of the [original BayDAG contribution](https://github.com/ActivitySim/activitysim/pull/657), the CDAP was extended to include a joint tour component (CDAP+J). This pull request adds required code needed to create a full estimation data bundle and code to estimate the CDAP+J model. The code to fully estimate the CDAP+J is not fully finished (SANDAG decided to just re-estimate the CDAP part), but is 90% of the way there. (The only remaining part is the final building of the joint utility equations, which is in the cdap_joint_tour_utility() method.) 


i-am-sijia (commented):
As described in the PR, the code changes here are not fully functioning for estimating the CDAP+J model.  The current estimation CI test is passing for regular CDAP, which means it's probably ok to merge this, and tackle the CDAP+J estimation in the future.

i-am-sijia (approved):

PR #771, opened 2023-12-18
BayDAG Contribution #5: Joint Tour Frequency and Composition Estimation

dhensle:
ActivitySim is very particular about what the index is named for the alternatives file when trying to create estimation data bundles.  (It comes back to this code [here](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/util/estimation.py#L346)). This fix is required in order for this model to run successfully in estimation mode. The joint tour frequency and composition model was included as part of the [original BayDAG contribution](https://github.com/ActivitySim/activitysim/pull/657).

### DIFF
--- a/activitysim/abm/models/joint_tour_frequency_composition.py
+++ b/activitysim/abm/models/joint_tour_frequency_composition.py
@@ -55,9 +55,11 @@ def joint_tour_frequency_composition(
             model_settings_file_name,
         )
 
+    # FIXME setting index as "alt" causes crash in estimation mode...
     alts = simulate.read_model_alts(
-        state, "joint_tour_frequency_composition_alternatives.csv", set_index="alt"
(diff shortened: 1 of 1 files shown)


i-am-sijia:
> (It comes back to this code [here](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/util/estimation.py#L346)).

404 Error. @dhensle Can you please update the reference link?

dhensle:
In the old version of the code which is what the link points to, its [here](https://github.com/ActivitySim/activitysim/blob/v1.2.2/activitysim/abm/models/util/estimation.py#L346).  In our current version, it's [here](https://github.com/ActivitySim/activitysim/blob/94c4db853458619d88e10239b4005a966dabb576/activitysim/core/estimation.py#L373).

i-am-sijia (approved):

Issue #731, opened 2023-09-16
Estimation Mode Refinement


joecastiglione:
- Consolidated Improve Estimation Functionality (https://github.com/ActivitySim/activitysim/issues/737) with this issue
- Consolidated Complete Estimation Mode for Trip Models (https://github.com/ActivitySim/activitysim/issues/726) with this issue

**Estimation Mode Refinement**

The current Estimation Mode features of ActivitySim are very much built in the spirit of similar functionality in DaySim: updated observed (survey) data can be fed into the process, and the combination of ActivitySim and Larch can work together somewhat automatically to generate updated parameter estimates.  The tools that have been built allow Larch to construct a model that exactly mirrors the defined model in ActivitySim, re-estimate model parameters, and output ActivitySim coefficient files with new parameter estimates that can be used as a drop-in replacement for the existing coefficient files.  

However, this tight integration breaks down when the user wants to update not only the coefficient values but also the functional form of utility equations.  This leaves the user with two choices: (a) returning to ActivitySim for every tiny change to the specification files and re-running the entire estimation mode process, or (b) editing the utility equations in Larch while exploring different functional forms, and then needing to reconstruct matching specification files later in ActivitySim once the desired function form is selected.  The former solution is tedious and slow, while the latter solution is error prone and requires fairly expert level understanding of the usage of both ActivitySim and Larch.   

The goals of this task would be to more tightly integrate Larch and ActivitySim, to achieve (1) allowing users to move between these tools using a common utility specification format, (2) to speed up the generation of data to support revisions to utility functional forms, especially for large data bundles (i.e. destination and scheduling components, possibly by sampling), (3) to extend and enhance the documentation of the estimation process, and (4) improve error handling.

**Improve Estimation Functionality**

ActivitySim currently uses the Larch software to estimate models, which allows estimation results to be used directly by the simulation — dramatically reducing errors common in translating utility expressions into the ActivitySim specification. In version 1.5, numerous improvements are made to the estimation procedures including reducing the size of the estimation data bundles, increasing the speed at which they can be estimated, improving the reporting and error messaging capabilities of Larch, and improving the usability of the coefficient files created by the ActivitySim procedures. Further, an auditing will be done to confirm that the estimation procedures for each ActivitySim component are working as expected.

**Complete Estimation Mode for Trip Models**

Fully implement estimation mode for all submodels. Seem like we are going to get close in phase 5, but may not be totally complete.

Additional Description:

[...]

PR #773, opened 2023-12-18
BayDAG Contribution #7: Sampling in EDB for Location Choice

dhensle:
Currently, estimation mode will include every single alternative for the location choice models.  This can be very onerous for 2-zone systems (SANDAG's workplace location choice estimation ran took multiple hours to run).  The code contributed here allows for the user to specify a sample rate to be used in estimation mode.  There is an additional logic to ensure that the actual observed alternative is included in the sampled alternatives so that downstream models will be using the correct destination. Using this "shortcut" for SANDAG decreased the runtime for this model to ~10 mins instead of hours.

Note that sampling is not yet supported in the actual model estimation process in larch yet. (Something [proposed for future enhancements](https://github.com/ActivitySim/activitysim/issues/731).)  This code just allows the user to run estimation mode much faster so downstream model EDBs can be created in less time.

### DIFF
--- a/activitysim/abm/models/location_choice.py
+++ b/activitysim/abm/models/location_choice.py
@@ -141,16 +141,18 @@ def _location_sample(
 
     sample_size = model_settings.SAMPLE_SIZE
 
-    if state.settings.disable_destination_sampling or (
-        estimator and estimator.want_unsampled_alternatives
-    ):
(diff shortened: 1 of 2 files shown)


i-am-sijia (approved):

jpn--:
This non-trivial PR has not test at all to confirm it works.

Issue #10, opened 2014-12-17
Progress report and discussion topics for 12-19-14 meeting


fscottfoti:
### Infrastructure

* [Pull request](https://github.com/synthicity/urbansim/pull/134) for (YAML-based) general choice modeling in Urbansim
* [Defaults](https://github.com/synthicity/activitysim/tree/adding-defaults/activitysim/defaults) for some standard data sources and variable definitions that are usable for all regions
* A [notebook](http://nbviewer.ipython.org/github/synthicity/activitysim/blob/adding-defaults/notebooks/data_mover.ipynb) for moving data from the directory structure into a single HDF5 file
* A [notebook](http://nbviewer.ipython.org/github/synthicity/activitysim/blob/adding-defaults/notebooks/simulation.ipynb) for browsing the registered tables and running the example models
* A [pull request](https://github.com/synthicity/activitysim/pull/7) for integrating OMX Python support
* A [pull request](https://github.com/synthicity/activitysim/pull/9) to wrap a 2D matrix with a Skim object 

### Specific Models

* We now have auto ownership and workplace location choice basically working on MTC data - this last uses the OMX and Skim objects.  There are a few design decisions that can be made at this point.  Here is a *proposal* for how the models could work.
  * An [example](https://github.com/synthicity/activitysim/tree/adding-defaults/example) implementation directory (these are client specific)
  * The configuration is still a [csv file](https://github.com/synthicity/activitysim/blob/adding-defaults/example/configs/auto_ownership_coeffs.csv) with the same basic form as the current UPCs.
  * Note that this is different from the current YAML approach in UrbanSim and is discussed in this [issue](https://github.com/synthicity/activitysim/issues/5)  
  * In short models are **specified** in the csv file, and configured in Python code like [this](https://github.com/synthicity/activitysim/blob/adding-defaults/example/models.py#L18).  Transformations that are more complicated than those allowed by Patsy can be specified in the CSV file - we use `DataFrame.eval` or straight Python `eval` when there is an `@` as the first character.  Allows more flexibility in the csv than is provided in current yaml files.
  * Additionally, there are simulation settings [here](https://github.com/synthicity/activitysim/blob/adding-defaults/example/configs/settings.yaml)
[...]

PR #776, opened 2023-12-18
BayDAG Contribution #10: NMTF Person Available Periods

dhensle:
The number of time periods available to a person after mandatory and joint tours are scheduled has a large influence on the number of non-mandatory tours they take. The current implementation only has the ability to return the number of consecutive available time periods available for the entire day.  However, the number of available bins between, e.g. 10PM and 3AM does not have much impact on non-mandatory travel and can drown out the impact of availability in the model. This code allows the user to be able to specify the start and end period and whether to return the result for the number of consecutive bins available or the total.

Non-mandatory tour frequency was re-estimated for SANDAG ABM3 and found significant coefficients for both total and consecutive time availability using this new functionality. See example [preprocessor](https://github.com/SANDAG/ABM/blob/ABM3_develop/src/asim/configs/resident/non_mandatory_tour_frequency_annotate_persons_preprocessor.csv#L8) and [spec](https://github.com/SANDAG/ABM/blob/ABM3_develop/src/asim/configs/resident/non_mandatory_tour_frequency.csv#L86).

Required for SANDAG ABM3 production? -- yes

### DIFF
--- a/activitysim/abm/models/non_mandatory_tour_frequency.py
+++ b/activitysim/abm/models/non_mandatory_tour_frequency.py
@@ -11,7 +11,7 @@ import numpy as np
 import pandas as pd
 
 from activitysim.abm.models.util import annotate
-from activitysim.abm.models.util.overlap import person_max_window
+from activitysim.abm.models.util.overlap import person_max_window, person_available_periods
 from activitysim.abm.models.util.school_escort_tours_trips import (
     recompute_tour_count_statistics,
 )
@@ -24,6 +24,7 @@ from activitysim.core import (
     simulate,
     tracing,
     workflow,
+    annotate,
 )
 from activitysim.core.configuration.base import PreprocessorSettings, PydanticReadable
 from activitysim.core.configuration.logit import LogitComponentSettings
@@ -230,7 +231,11 @@ def non_mandatory_tour_frequency(
     # - preprocessor
     preprocessor_settings = model_settings.preprocessor
     if preprocessor_settings:
-        locals_dict = {"person_max_window": lambda x: person_max_window(state, x)}
+
+        locals_dict = {
+            "person_max_window": lambda x: person_max_window(state, x),
+            "person_available_periods": lambda x: person_available_periods(state, x),
+        }
 
         expressions.assign_columns(
             state,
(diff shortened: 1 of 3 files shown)


activitysim/abm/models/non_mandatory_tour_frequency.py
```diff
@@ -418,13 +428,14 @@ def non_mandatory_tour_frequency(
...
         # make sure they created the right tours
         survey_tours = estimation.manager.get_survey_table("tours").sort_index()
-        non_mandatory_survey_tours = survey_tours[
-            survey_tours.tour_category == "non_mandatory"
-        ]
-        assert len(non_mandatory_survey_tours) == len(non_mandatory_tours)
-        assert non_mandatory_survey_tours.index.equals(
-            non_mandatory_tours.sort_index().index
-        )
+        # FIXME below check needs to remove the pure-escort tours from the survey tours table
```
i-am-sijia: Would it be better to optionally skip this section of code instead of simply commenting them out? What is the use case for commenting them out?

activitysim/abm/models/non_mandatory_tour_frequency.py
```diff
@@ -324,6 +331,9 @@ def non_mandatory_tour_frequency(
 
         choices_list.append(choices)
 
+    # FIXME only want to keep actual purposes, adding cols in alts will mess this up
+    # this is complicated by canonical_ids calculated based on alts if not specified explicitly
+    # thus, adding column to input alts will change IDs and break estimation mode....
```
i-am-sijia: I believe this is fixed given the last discussion and code change here -> https://github.com/ActivitySim/activitysim/pull/661

i-am-sijia (changes requested):

activitysim/abm/models/non_mandatory_tour_frequency.py
dhensle: Updated this code segment with code that addresses the FIXME comment and turned the checks back on.

activitysim/abm/models/non_mandatory_tour_frequency.py
dhensle: Agreed.  Removed outdated comment.

i-am-sijia (approved):

PR #777, opened 2023-12-18
BayDAG Contribution #11: School Escorting Estimation Updates

dhensle:
Most changes in this PR are needed to avoid crash if estimation run actually had no school escorting in the input data.  Empty school escorting tables are not handled appropriately in the current implementation.  This would arise when running estimation mode with a small enough sample that school escorting was not included in the estimation data.

Additional fixes in this PR fix a case where the "gender weight" optional setting used to rank chauffeurs was incorrectly using the person weight setting instead. It also appropriately names the alternative column for estimation mode (see #771).

### DIFF
--- a/activitysim/abm/models/school_escorting.py
+++ b/activitysim/abm/models/school_escorting.py
@@ -58,7 +58,14 @@ def determine_escorting_participants(
         & (persons.cdap_activity == "M")
     ]
     households_with_escortees = escortees["household_id"]
-    choosers = choosers[choosers.index.isin(households_with_escortees)]
+    if len(households_with_escortees) == 0:
+        logger.warning("No households with escortees found!")
+    else:
+        tot_households = len(choosers)
+        choosers = choosers[choosers.index.isin(households_with_escortees)]
+        logger.info(
+            f"Proceeding with {len(choosers)} households with escortees out of {tot_households} total households"
+        )
 
     # can specify different weights to determine chaperones
     persontype_weight = model_settings.PERSON_WEIGHT
@@ -676,6 +683,7 @@ def school_escorting(
             "destination",
             "escort_participants",
             "chauf_tour_id",
+            "primary_purpose",
         ]
         school_escort_trips = pd.DataFrame(columns=trip_cols)
 
--- a/activitysim/abm/models/school_escorting.py
+++ b/activitysim/abm/models/school_escorting.py
@@ -486,7 +486,9 @@ def school_escorting(
     for stage_num, stage in enumerate(school_escorting_stages):
         stage_trace_label = trace_label + "_" + stage
         estimator = estimation.manager.begin_estimation(
-            state, model_name="school_escorting_" + stage, bundle_name="school_escorting"
+            state,
+            model_name="school_escorting_" + stage,
+            bundle_name="school_escorting",
         )
 
         model_spec_raw = simulate.read_model_spec(
--- a/activitysim/abm/models/school_escorting.py
+++ b/activitysim/abm/models/school_escorting.py
@@ -140,7 +140,7 @@ def add_prev_choices_to_choosers(
             stage_alts,
             how="left",
             left_on=escorting_choice,
-            right_on=stage_alts.index.name,
+            right_index=True,
         )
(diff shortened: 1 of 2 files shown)


activitysim/abm/models/school_escorting.py
```diff
@@ -478,7 +486,13 @@ def school_escorting(
...
         stage_trace_label = trace_label + "_" + stage
         estimator = estimation.manager.begin_estimation(
-            state, "school_escorting_" + stage
+            state,
+            model_name="school_escorting_" + stage,
+            bundle_name="school_escorting",
+        )
+
+        model_spec_raw = state.filesystem.read_model_spec(
+            file_name=getattr(model_settings, stage.upper() + "_SPEC")
```
i-am-sijia: duplicative code, can be deleted.

activitysim/abm/models/school_escorting.py
```diff
@@ -460,7 +471,11 @@ def school_escorting(
 
     trace_hh_id = state.settings.trace_hh_id
 
-    alts = simulate.read_model_alts(state, model_settings.ALTS, set_index="Alt")
+    # FIXME setting index as "Alt" causes crash in estimation mode...
+    # happens in joint_tour_frequency_composition too!
+    # alts = simulate.read_model_alts(state, model_settings.ALTS, set_index="Alt")
+    alts = simulate.read_model_alts(state, model_settings.ALTS, set_index=None)
+    alts.index = alts["Alt"].values
```
i-am-sijia, line 478: I noticed all other models' alternatives files have the alternative column named and **hardcoded** in source code as `alt`, but for school escorting, it is `Alt`.  It's probably better to keep the column name consistent as `alt`, even though we are not addressing the "hardcoded" part at this moment.

activitysim/abm/models/school_escorting.py
```diff
@@ -533,9 +554,26 @@ def school_escorting(
...
+            #  shuold we do it here or have interaction_simulate do it?
+            # chooser index must be duplicated in column or it will be omitted from interaction_dataset
+            # estimation requires that chooser_id is either in index or a column of interaction_dataset
+            # so it can be reformatted (melted) and indexed by chooser_id and alt_id
+            assert choosers.index.name == "household_id"
+            assert "household_id" not in choosers.columns
+            choosers["household_id"] = choosers.index
+
+            # FIXME set_alt_id - do we need this for interaction_simulate estimation bundle tables?
+            estimator.set_alt_id("alt_id")
```
i-am-sijia, line 570: The question in the FIXME - does it remain as a question?

i-am-sijia (changes requested):
Overall the code changes look fine, I have a few comments for you to take a look.

activitysim/abm/models/school_escorting.py
dhensle: I think we should treat this issue separately from this PR.  If we change it here, it will have implications for people who are already using the model (i.e. they would have to change their alternatives file). This is larger than simply changing it here as the Alt name is coded in a couple of places in this file and I would like to keep this PR focused on the BayDAG contributions.  Thus, I think we should track it in a separate and stand-alone issue. Thoughts?

activitysim/abm/models/school_escorting.py
dhensle: good catch, fixed.

activitysim/abm/models/school_escorting.py
dhensle: Yes, this was copied directly from the non-mandatory tour frequency model since I am keeping this consistent with what was done for that model (see [here](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/non_mandatory_tour_frequency.py#L297)).  My guess is that it has been around since estimation mode was first developed. This note here is to help us keep track of this issue across all locations in the code. (It shows up in the vehicle type choice model as well.)

activitysim/abm/models/school_escorting.py
jpn--: Created #846

jpn-- (approved):

PR #781, opened 2023-12-18
BayDAG Contribution #15: Larch Interface for Added Models

dhensle:
This PR creates an entry point into Larch for added newly models "simple simulate" models, namely work from home, telecommute frequency, transit pass subsidy, and transit pass ownership models. These models can now be loaded into Larch via an estimation notebook calling the corresponding functions.


i-am-sijia (approved):
Code changes looks fine. With caveat that there is no test to confirm it works.

Issue #673, opened 2023-05-08
Optimize Data Type Usage


i-am-sijia:
## Background
In the memory profiling work conducted by WSP during Phase 8 Interim, we have identified that memory usage of the ActivitySim model is high when large “choosers” tables with inefficient data types are created. For example, in the work tour scheduling step in the ARC run with a 25 percent sample, the choosers table has 540 million rows and 80 columns – a memory footprint of about 250 GB. Among the 80 columns, there are five string variables including tour purpose (e.g., “work”, “school”), tour category (e.g., “mandatory”), and time-period (e.g., “AM”); each takes about 32 GB of RAM. If we change string variables into something like enums using int8 data types, the memory footprint of this chooser table could be reduced from 250 GB to 102 GB. Many of the other columns unnecessarily use memory-intensive data types like float64 and int64. A logical next step is to optimize the data types used in ActivitySim, as part of the Phase 8 development.

### String variables

We looked into two alternatives to optimizing string variables:

- IntEnum. Presented at the [June 27, 2023 Meeting](https://github.com/ActivitySim/activitysim/wiki/Project-Meeting-2023.06.27).
- Pandas Categorical data type. Presented at the [July 18, 2023 Meeting](https://github.com/ActivitySim/activitysim/wiki/Project-Meeting-2023.07.18).

The table below recaps the pros and cons of the two alternatives:

![image](https://github.com/ActivitySim/activitysim/assets/28812722/8ceb077f-eb9b-46cc-9ea4-20620a461c3f)

In consultation with the consortium members and the bench contractors, we decided to implement pandas Categorical for converting string variables, mainly because the level of effort is lower and it will keep the backward compatibility. We will modify the ActivitySim source code so that when a string variable is created, we convert it to pandas categorical.

### Numeric variables

Numeric variables are created and used in the following sources:
1. Input data
2. pre- and post- annotation of each ActivitySim sub-model
3. .py source code

For the numeric variables in the input data, the user can define their data types in the settings.yaml, see example here. For the numeric variables created in the annotations and source codes, we can create a function that downcasts them based on the value ranges of the variables.

### Overview
The string to pandas categorical conversion shall happen under the hood, in the ActivitySim source code, and it should require minimum work for users to implement their models with this change. The downcasting of numeric variables is implemented as an option that users can turn on and off.

Relevant discussions/presentations can be found at:
[...]

Issue #766, opened 2023-12-15
Utility terms that compare pandas categorical variable to strings are not evaluated correctly with Sharrow


i-am-sijia:
**Describe the bug**
After implementing the string to pandas categorical conversion, some of our current CI tests failed. They all had Sharrow turned on and set to `test` mode. The utility calculated with and without Sharrow are different.

```python
[48:39.10] INFO: completed flow_LQLDEWSFEGQ5W2NJNONCWPNFSCB7O5RD.load in 0:00:18.710844 stop_frequency.work.simple_simulate.eval_mnl
[48:39.10] INFO: completed apply_flow in 0:00:21.651810 
[48:39.10] INFO: elapsed time sharrow flow 0:00:21.659376 stop_frequency.work.simple_simulate.eval_mnl
[48:39.31] INFO: elapsed time simple flow 0:00:00.207622 stop_frequency.work.simple_simulate.eval_mnl.eval_utils

Not equal to tolerance rtol=0.01, atol=0
utility not aligned
Mismatched elements: 132 / 144 (91.7%)
Max absolute difference: 1998.00011762
Max relative difference: 1729.2712081
 x: array([[    0.     , -1000.9582 , -1002.2882 , -1002.6522 , -1001.3462 ,
        -2000.7913 , -2002.1212 , -2002.4852 , -1003.1262 , -2002.5713 ,
        -2003.9012 , -2003.5703 , -1004.4472 , -2003.8922 , -2004.5272 ,...
 y: array([[ 0.000000e+00, -1.958200e+00, -3.288200e+00, -3.652200e+00,
        -2.346200e+00, -2.791200e+00, -4.121200e+00, -4.485200e+00,
        -4.126200e+00, -4.571200e+00, -5.901200e+00, -5.570200e+00,...
big problem: 132 missed close values out of 144 (91.67%)
sh_util.shape=(9, 16)
(array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1,
       1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
       2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4, 4, 4,
[...]

PR #782, opened 2023-12-28
Data Type Optimization

i-am-sijia:
PR for https://github.com/ActivitySim/activitysim/issues/673


activitysim/core/util.py
```diff
@@ -323,6 +326,8 @@ def assign_in_place(df, df2):
         assignment left-hand-side (dest)
     df2: pd.DataFrame
         assignment right-hand-side (source)
+    downcast_float: bool
```
jpn--, line 329: include downcast_int also.

activitysim/core/util.py
```diff
@@ -371,6 +379,79 @@ def assign_in_place(df, df2):
...
+                continue
+            else:
+                df[col] = pd.to_numeric(df[col], downcast="integer")
+                continue
+            # there are calcualtions in asim that expect results in negative values
+            # operations on two unsigned types will not produce negative values
+            # if df[col].min() >= 0:
+            #     df[col] = pd.to_numeric(df[col], downcast='unsigned')
+            # else:
+            #     df[col] = pd.to_numeric(df[col], downcast='integer')
```
jpn--: Looks like we're circumventing this by just downcasting to signed integer regardless.  We don't need this commented _code_ and it can come out, but do please leave/update the contextual comment above it about this topic, so some poor soul doesn't come back in the future and try to save a few bits of RAM by doing this.

activitysim/abm/models/vehicle_type_choice.py
```diff
@@ -231,7 +231,9 @@ def construct_model_alternatives(
         else:
             # eliminate alternatives if no vehicle type data
             alts_wide = alts_wide[alts_wide._merge != "left_only"]
-        alts_wide.drop(columns="_merge", inplace=True)
+        alts_wide.drop(
+            columns="_merge", inplace=True
+        )  # if this happens, alt_wide length will not match alt_long
```
jpn--: Considering the comment, is this then a bug that needs to be addressed?

activitysim/abm/models/util/test/test_mandatory_tour_frequency.py
```diff
@@ -56,10 +56,11 @@ def test_mtf():
     pdt.assert_series_equal(
         mandatory_tours.person_id,
         pd.Series([10, 20, 20, 30, 30, 40, 40], index=idx, name="person_id"),
+        check_dtype=False,
     )
 
     pdt.assert_series_equal(
-        mandatory_tours.tour_type,
+        mandatory_tours.tour_type.astype(str),
```
jpn--: instead of dumping to `str`, perhaps we should have the test verify the categorical data type here?

activitysim/core/util.py
```diff
@@ -371,6 +379,79 @@ def assign_in_place(df, df2):
 
     df[new_columns] = df2[new_columns]
 
+    for c in new_columns:
+        if pd.api.types.is_object_dtype(df[c]):
+            df[c] = df[c].astype("category")
+
+    auto_opt_pd_dtypes(df, downcast_int, downcast_float, inplace=True)
```
jpn--, line 385: By running this at the end of the assign-in-place func, it seems like we're first creating a large, inefficient dataframe with poorly optimized data types, and then trying to compress it down by using more efficient types.  In many cases this is probably fine, the data types are not really that big and it's keeping the size down on long-term stored data, but in corner cases this may make the memory usage worse instead of better

activitysim/examples/placeholder_psrc/configs/stop_frequency_annotate_tours_preprocessor.csv
```diff
@@ -45,3 +45,6 @@ AccesibilityADestination off-peak transit,_dest_trOpRetail,"reindex(accessibilit
 AccesibilityAtDestination if transit,pracc,"pracc.where(~tour_mode_is_transit, _dest_trPkRetail.where(_tour_starts_in_peak, _dest_trOpRetail))"
 AccesibilityAtDestination if non_motorized,pracc,"pracc.where(~tour_mode_is_non_motorized, reindex(accessibility.nmRetail, df.destination))"
 ,destination_area_type,"reindex(land_use.area_type, df.destination)"
+,is_mixed,df.composition=='mixed'
```
jpn--, line 48: general comment: the changes we are making in all these model spec and preprocessor files, are they *required* or just helpful?  We don't want to have required changes that injure backwards compatibility.

jpn-- (changes requested):
This is a partial review, there's a lot to go through here and I'm not done but want to share my comments so far instead of holding them all until I'm done.

activitysim/core/util.py
i-am-sijia: Yep, I deleted them. I should probably open an issue about the negative values (although it's not really an issue), since it impacts on the data type to use. It would be good to remember this if/when we do the data model.

activitysim/abm/models/util/test/test_mandatory_tour_frequency.py
i-am-sijia: I see, they are not pd.cat in this unit test any more. This test calls the method in the util/tour_frequency.py. Initially I implemented the pd.cat conversion in the tour_frequency.py, with hardcoded categories. More recently, I moved the pd.cat conversion to the actual tour frequency models themselves, like mandatory_tour_frequency.py, so that the categories can be automatically set based on the UECs instead of hardcode in .py.

activitysim/core/util.py
i-am-sijia: Correct, auto_opt_pd_dtypes() compresses variables after they are created in, e.g., inputs, annotations, UECs. It's a mask and shortcut rather than changing how activitysim creates variables, i.e., eval().  For this instance, are you suggesting applying auto_opt_pd_dtypes() to df2, rather than df?  When you say "in corner cases", do you mean how pandas converts data types internally and it's possible for it to consume more memory?

activitysim/examples/placeholder_psrc/configs/stop_frequency_annotate_tours_preprocessor.csv
i-am-sijia: These are required in sharrow mode. It is related to [issue 766](https://github.com/ActivitySim/activitysim/issues/766). I noticed that if a pandas categorical variable is compared to a string, sharrow seems to evaluate it wrong.

activitysim/abm/models/vehicle_type_choice.py
i-am-sijia: Opened issue: https://github.com/ActivitySim/activitysim/issues/789

i-am-sijia: I moved this comment up two lines.

activitysim/core/util.py
jpn--: When I wrote this I was talking here about how the basic `eval` code creates object type variables, and wouldn't it be better if it didn't.  But I think you're right not to touch that yet, it's likely to get messy and be not-so backwards compatible, which we don't want to do (yet)

activitysim/examples/placeholder_psrc/configs/stop_frequency_annotate_tours_preprocessor.csv
jpn--: ok, I have added 766 to the Phase 9 priority list.

i-am-sijia:
@jpn-- seems the change of the black version has caused the linting test to fail. I can update my local black and reformat if you are ok with that.

jpn--:
> @jpn-- seems the change of the black version has caused the linting test to fail. I can update my local black and reformat if you are ok with that.

No, I made a mistake and clicked the wrong button to merge code into develop I wasn't intending to.   I rolled it back, should be better if you just merge in the latest changes from the develop branch.

jpn-- (approved):

Issue #789, opened 2024-02-01
Vehicle type model `REQUIRE_DATA_FOR_ALL_ALTS` needs to be True


i-am-sijia:
**Describe the bug**

The vehicle type choice model creates alts_wide and alts_long tables, and they need to be in the same length and order, because this line of code assumes that for the final mapping of choices:

https://github.com/ActivitySim/activitysim/blob/a8e755f96d0e32633a6d3657c4878e3b6a37e59a/activitysim/abm/models/vehicle_type_choice.py#L426

However, the section of code below drops alternatives from alts_wide if there are alternatives that do not exist in the input `VEHICLE_TYPE_DATA_FILE` and if the optional `REQUIRE_DATA_FOR_ALL_ALTS` does not get defined or is set to False.  

https://github.com/ActivitySim/activitysim/blob/a8e755f96d0e32633a6d3657c4878e3b6a37e59a/activitysim/abm/models/vehicle_type_choice.py#L217-L224

This will make the alts_wide inconsistent with the alts_long, and make the final mapping of choices incorrect.

The model will still run, but the result will be incorrect.

**To Reproduce**

1. Delete the first row of the [config/vehicle_type_data.csv](https://github.com/ActivitySim/activitysim/blob/a8e755f96d0e32633a6d3657c4878e3b6a37e59a/activitysim/examples/prototype_mtc_extended/configs/vehicle_type_data.csv#L2)
2. Run the existing prototype_mtc_extended model
3. The model will run, but it creates different result for vehicle type

**Expected behavior**

The model should fail if there are alternatives that do not exist in the vehicle type data. `REQUIRE_DATA_FOR_ALL_ALTS` should not be optional or and should be True. The IF condition on Line 217 can be replaced by the assert.

dhensle:
Good find @i-am-sijia.  I believe I have a fix (#790) in place to also filter alts_long if alts_wide gets filtered. I believe this means the `REQUIRE_DATA_FOR_ALL_ALTS` option is usable.

i-am-sijia:
Ok, let's add `REQUIRE_DATA_FOR_ALL_ALTS` to the model documentation, and update the following blurb to remove "must be included in the file" -
https://github.com/ActivitySim/activitysim/blob/a8e755f96d0e32633a6d3657c4878e3b6a37e59a/docs/models.rst#L448-L451
[...]

PR #790, opened 2024-02-02
Vehicle Type Alts Filtering Bug Fix

dhensle:
PR for #789

### DIFF
--- a/activitysim/abm/models/util/test/test_vehicle_type_alts.py
+++ b/activitysim/abm/models/util/test/test_vehicle_type_alts.py
@@ -1,0 +1,57 @@
+# ActivitySim
+# See full license in LICENSE.txt.
+
+import pandas as pd
+import pandas.testing as pdt
+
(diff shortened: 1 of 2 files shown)


jpn--:
When we fix a bug, we need to have a test that proves it's fixed (and so it stays fixed).

i-am-sijia:
Need to update documentation:
> Ok, let's add `REQUIRE_DATA_FOR_ALL_ALTS` to the model documentation, and update the following blurb to remove "must be included in the file" -
> 
> https://github.com/ActivitySim/activitysim/blob/a8e755f96d0e32633a6d3657c4878e3b6a37e59a/docs/models.rst#L448-L451

jpn--:
I opened a PR #796 to bring ActivitySim's config settings up-to-date with Pydantic 2.0, which should address your most recent test failure (at least in part).

dhensle:
Thanks! The test was passing locally, so I expect the update should fix the test completely.

PR #796, opened 2024-02-08
Pydantic 2

### LINKED ISSUE #795: Compatibility with Pydantic 2.x

jpn--:
Update ActivitySim to use Pydantic version 2 for settings validation.  There are small changes in the pydantic API to accommodate.

It should be a smidge faster, although settings validation is such a small part of overall runtime that it's likely not noticeable (unless we eventually use Pydantic for more).  But more importantly, this is the most up-to-date version of Pydantic so compatibility with that tool and others that rely on it will be more robust. 

This PR closes #795

### DIFF
--- a/.github/workflows/core_tests.yml
+++ b/.github/workflows/core_tests.yml
@@ -49,7 +49,7 @@ jobs:
           mamba env update -n asim-test -f conda-environments/github-actions-tests.yml
           mamba install --yes \
           	"psutil=5.9.5" \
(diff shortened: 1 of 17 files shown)


dhensle (commented):
I think we can remove this commented out reference to the old parse_obj() call: https://github.com/camsys/activitysim/blob/pydantic-2/activitysim/core/mp_tasks.py#L890

Other than that, I think it looks good.

PR #798, opened 2024-02-13
Alt col name bug fix for option to handle larger dataset for location models 

bwentl:
This is to fix bug related to a hard-coded variable name in #687 when building the estimation data in chunks.

The previous pull request uses exclusively person_id, which means it will only work for mandatory destination EDBs. This change will mean that tour and trip destination EDBs will also work. This fix relies on the ID column inside the alternative data for EDB being the first column, which currently is the case.


jpn-- (approved):

jpn--:
This looks fine.  We need a more robust estimation mode system with rigorous testing to catch these problems, but this is a reasonable patch for now.  Will merge once we confirm none of the existing tests are broken (which seems quite unlikely but just in case...)  Thanks!

PR #808, opened 2024-02-20
Fix at-work frequency

jpn--:
Fixes a bug in at-work frequency, if there are no work tours in a chunk.


i-am-sijia (approved):
Thank you for catching it!

PR #812, opened 2024-02-22
Infrastructure Updates

jpn--:
This PR updates some of the packaging infrastructure.

- `setup.cfg` and `setup.py` are completely removed in favor of moving all setup into the `pyproject.toml` file, consistent with latest Python packaging practices.
- The minimum Python version is pushed to 3.10 (primarily to allow type annotations in `X | Y` format, for ease of use with Pydantic).
- The `sparse` package is added as a required dependency.  This package is required for 2-zone models, and adds very little overhead to the install even for 1-zone models; it is easier to install it all the time rather than rely on users to know if it would be needed or not.


i-am-sijia:
Change the following to `destination_dir: ${{ github.ref_name }}` ?

https://github.com/ActivitySim/activitysim/blob/5e6795ed614dd87e268ab889304ebdb1453aa6dd/.github/workflows/core_tests.yml#L564

i-am-sijia (approved):
I tested the installation on my windows laptop, it worked fine and ran the example mtc model. I had one question in my other comment.

jpn--:
> Change the following to `destination_dir: ${{ github.ref_name }}` ?
> 
> https://github.com/ActivitySim/activitysim/blob/5e6795ed614dd87e268ab889304ebdb1453aa6dd/.github/workflows/core_tests.yml#L564

I'm leaving this as is -- every time we have a PR against `main`, if it passes the foundational tests then we build and push the documentation to the "develop" subdirectory where our in-development docs live.  There are other actions that push docs for specific releases etc.

PR #824, opened 2024-02-29
Common settings for fastmath, sharrow_skip, and other compute controls across components

jpn--:
The reason why the school escort model was failing with sharrow relates to the `fastmath` setting.

Part of how sharrow accelerates code evaluation is by turning on the "[fastmath](https://numba.readthedocs.io/en/stable/user/performance-tips.html#fastmath)" compiler optimizations in Numba.  Using this setting has numerous effects, but two are salient here: 

1. It makes the code run faster 
2. It can cause elementary operations to give unexpected results where one or more operands is `NaN`, e.g. normally evaluating `x < 0` should result in `False` when `x` is `NaN` but with `fastmath` turned on the result might be `True` instead; the actual result of such an operation is not defined and might be hardware dependent.

We do not want to simply turn off `fastmath`, as the speed we get from (1) is desirable and the problems created by (2) don't actually occur in most models, as most data columns in our tables don't include missing values.  But the school escort model DOES include missing values (e.g. the age of the 3rd child in a household with less than 3 children).  

This PR introduces a `sharrow_fastmath` setting for all relevant components, with `True` as the default for nearly all Sharrow-compiled model components, but adds a mechanism to turn it off for individual components as needed.  The school escort model has the default for this setting changed to `False`.

### DIFF
--- a/activitysim/core/configuration/logit.py
+++ b/activitysim/core/configuration/logit.py
@@ -8,7 +8,7 @@ import pydantic
 from pydantic import BaseModel as PydanticBase
 from pydantic import model_validator, validator
 
-from activitysim.core.configuration.base import PreprocessorSettings, PydanticSharrow
+from activitysim.core.configuration.base import PreprocessorSettings, PydanticCompute
 
 
 class LogitNestSpec(PydanticBase):
@@ -45,7 +45,7 @@ class LogitNestSpec(PydanticBase):
         return coefficient_value
 
 
-class BaseLogitComponentSettings(PydanticSharrow):
+class BaseLogitComponentSettings(PydanticCompute):
     """
     Base configuration class for components that are logit models.
 
--- a/activitysim/core/configuration/logit.py
+++ b/activitysim/core/configuration/logit.py
@@ -77,33 +77,33 @@ class BaseLogitComponentSettings(PydanticSharrow):
     CONSTANTS: dict[str, Any] = {}
     """Named constants usable in the utility expressions."""
 
-    # sharrow_skip is deprecated in factor of sharrow_settings.skip
+    # sharrow_skip is deprecated in factor of compute_settings.sharrow_skip
     @model_validator(mode="before")
     @classmethod
     def update_sharrow_skip(cls, data: Any) -> Any:
         if isinstance(data, dict):
             if "sharrow_skip" in data:
-                if "sharrow_settings" not in data:
+                if "compute_settings" not in data:
                     # move to new format
-                    data["sharrow_settings"] = {"skip": data["sharrow_skip"]}
+                    data["compute_settings"] = {"sharrow_skip": data["sharrow_skip"]}
                     del data["sharrow_skip"]
                     warnings.warn(
-                        "sharrow_skip is deprecated in favor of sharrow_settings.skip",
+                        "sharrow_skip is deprecated in favor of compute_settings.sharrow_skip",
                         DeprecationWarning,
                     )
                 elif (
-                    isinstance(data["sharrow_settings"], dict)
-                    and "skip" not in data["sharrow_settings"]
+                    isinstance(data["compute_settings"], dict)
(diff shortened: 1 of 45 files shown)


dhensle:
Just wanted to make sure you knew that the string removal work was performed already here: https://github.com/ActivitySim/activitysim-prototype-mtc/pull/4

activitysim/core/configuration/logit.py
```diff
@@ -78,6 +78,17 @@ class BaseLogitComponentSettings(PydanticReadable):
     sharrow_skip: bool = False
     """Skip sharrow when evaluating this component."""
 
+    sharrow_fastmath: bool = True
```
dhensle: Do we want a top level setting that we can turn it off for all models?  If I understand the code right now, to turn it off we would have to add this to all model yamls, right?

dhensle (changes requested):
There are a number of submodels which do not include this setting, e.g. destination models, scheduling, transit pass subsidy, etc.   Those also need to be passed with the fastmath argument, especially since their yaml files will now accept the setting since the BaseLogitComponentSettings was updated.

jpn--:
@dhensle you make a good point.  Also, while I'm pushing fastmath through all the components, I realized its convenient to wrap all component level sharrow controls together (i.e., also `skip`) to ensure a consistent user interface.

activitysim/core/configuration/logit.py
jpn--: Yes, to turn fastmath off globally you'd need to turn it off in every component.  We can consider adding that in the future if it is desirable.

jpn--:
@dhensle this PR is ready for your re-review.  I believe I addressed your concerns, and then some.

activitysim/core/interaction_simulate.py
```diff
@@ -82,10 +84,12 @@ def eval_interaction_utilities(
...
 
     sharrow_enabled = state.settings.sharrow
-
-    if locals_d is not None and locals_d.get("_sharrow_skip", False):
+    if sharrow_settings is not None and sharrow_settings.skip:
         sharrow_enabled = False
 
+    # if locals_d is not None and locals_d.get("_sharrow_skip", False):
+    #     sharrow_enabled = False
+
```
dhensle: We can remove this commented code.

activitysim/core/interaction_simulate.py
```diff
@@ -691,11 +697,14 @@ def _interaction_simulate(
...
+    if sharrow_settings is not None and sharrow_settings.skip:
+        sharrow_enabled = False
+    else:
+        sharrow_enabled = state.settings.sharrow
     interaction_utilities = None
 
-    if locals_d is not None and locals_d.get("_sharrow_skip", False):
-        sharrow_enabled = False
+    # if locals_d is not None and locals_d.get("_sharrow_skip", False):
+    #     sharrow_enabled = False
```
dhensle: same here. cleanup comments

jpn--:
In thinking about the solution for #842, I realized we probably want these controls to be more general that "sharrow_settings", something more like "compute_settings".  The reason we are seeing the overflow in #842 seems to be because we are using bottleneck for the computations, but if we activate `compute.use_numexpr` instead the problem goes away.  If we make these controls more general, we can trigger this pandas setting here as well.  This allows us to change compute settings in a consistent component-specific manner for non-sharrow compute.  For now, that includes settings the pandas options `compute.use_bottleneck`, `compute.use_numexpr`, and `compute.use_numba` (see [pandas.set_option](https://pandas.pydata.org/docs/reference/api/pandas.set_option.html)).

jpn--:
@dhensle Will you have time to finish a (re)review of this PR this morning so we can merge it?  It is now blocking progress on #833 as we want to coordinate the settings.

dhensle (changes requested):
I think you just missed the transit_pass_subsidy model.  Other than that I think it looks good.

Issue #822, opened 2024-02-27
School escort model runtime


dhensle:
School escorting model has a long tail when running with larger samples: 

![Image](https://github.com/ActivitySim/activitysim/assets/51132108/676d0dd9-1b15-4809-a23c-b1db55cbbc53)

Doing a pyinstrument trace shows 

![Image](https://github.com/ActivitySim/activitysim/assets/51132108/192266d6-0743-4b62-8881-a2149fe24bd5)

Which shows the following functions need to be optimized:

- [ ] [create_child_escorting_stops()](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/util/school_escort_tours_trips.py#L161)
- [ ] [create_chauf_escort_trips()](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/util/school_escort_tours_trips.py#L79)
- [ ] [create_bundle_attributes](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/models/school_escorting.py#L151)

All of these functions are called from a pd.DataFrame.apply() method.

PR #828, opened 2024-03-05
School Escorting Runtime Optimization

dhensle:
PR for https://github.com/ActivitySim/activitysim/issues/822


activitysim/abm/models/util/test/test_school_escorting_utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def test_create_bundle_attributes():
+    data_dir = os.path.join(os.path.dirname(__file__), "data")
+
+    inbound_input = pd.read_pickle(
+        os.path.join(data_dir, "create_bundle_attributes_inbound__input.pkl")
+    )
+    inbound_expected = pd.read_pickle(
+        os.path.join(data_dir, "create_bundle_attributes_inbound__output.pkl")
```
i-am-sijia, line 23: Can we use parquet instead of pickle for test data?  Pickle is proven to have compatibility issues when changing version of Python or dependency (like pandas).

dhensle: As discussed during the meeting today, the parquet format does not preserve the dtype of the column like Pickle does. This is relevant in particular for the school escorting model because some columns contain arrays of values (like escorting participants or the school destinations of the escort bundle for example).  These arrays are not stored appropriately by the parquet format. 

The long-term solution agreed upon was to push this issue of parquet compatability (which I think happens in a couple of other places besides the school escort model) to the data model.

In order to make sure that this isn't an issue that was going to pop up immediately, the request was to test with pandas 2.  I can confirm that this unit test passes with the latest version of Pandas v2.2.2.

i-am-sijia (approved):

jpn-- (approved):

PR #834, opened 2024-03-19
BayDAG All Contributions

dhensle:
One master PR that merges in the other BayDAG pull requests.  Used to test the SANDAG ABM3 example.

### DIFF
--- a/activitysim/abm/models/auto_ownership.py
+++ b/activitysim/abm/models/auto_ownership.py
@@ -36,6 +36,7 @@ def auto_ownership_simulate(
     state: workflow.State,
     households: pd.DataFrame,
     households_merged: pd.DataFrame,
+    # FIXME: persons_merged not used but included, see #853
     persons_merged: pd.DataFrame,
     model_settings: AutoOwnershipSettings | None = None,
     model_settings_file_name: str = "auto_ownership.yaml",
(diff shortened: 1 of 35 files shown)


activitysim/abm/models/auto_ownership.py
```diff
@@ -36,6 +36,7 @@ def auto_ownership_simulate(
     state: workflow.State,
     households: pd.DataFrame,
     households_merged: pd.DataFrame,
+    persons_merged: pd.DataFrame,
```
jpn--, line 40: Seems like `persons_merged` isn't used in auto ownership, why are we adding it here?

jpn-- (commented):
Only one minor change (and updating to clear the very simple conflicts against now-current main), then this looks clear to merge.

activitysim/abm/models/auto_ownership.py
dhensle: It is actually used in the preprocessor settings in SANDAG's ABM3 [here](https://github.com/ActivitySim/sandag-abm3-example/blob/main/configs/resident/auto_ownership_preprocessor.csv).  The issue is that the persons_merged table needs to include the disaggregate accessibility outputs.  Including the persons_merged table here in the call forces the person_merged table to be re-merged to include the disaggregate accessibility columns.  This wouldn't ordinarily be needed, but for ABM3 the auto_ownership model is the first model in the list so the persons_merged table wasn't yet updated.  

I am open to a different solution here if you have a better way to handle this.

jpn--: Can we just add a comment in the code next to it that it's needed in the args list for the side effects?

I'm not a fan of the whole side effects system, but I don't have a better solution here that's easy.

dhensle: Made issue #853 and referenced it in the code comment.

PR #839, opened 2024-03-27
fix bug in interaction_simulate

jpn--:
We recently fixed a problem where sharrow was holding on to data references, sometimes preventing memory from being freed.

This introduced a bug in the sharrow-test error handler of the interaction simulate code, where sharrow re-runs data if it does not match utility values with the original non-sharrow code.  This PR fixes the bug.

### DIFF
--- a/activitysim/core/interaction_simulate.py
+++ b/activitysim/core/interaction_simulate.py
@@ -565,9 +565,7 @@ def eval_interaction_utilities(
                     retrace_eval_data_ = pd.concat(retrace_eval_data, axis=1)
                     retrace_eval_parts_ = pd.concat(retrace_eval_parts, axis=1)
 
-                    re_sh_flow_load = sh_flow.load(
-                        dtype=np.float32,
-                    )
+                    re_sh_flow_load = sh_flow.load(sh_tree, dtype=np.float32)
                     re_sh_flow_load_ = re_sh_flow_load[re_trace]
 
                     look_for_problems_here = np.where(


jpn--:
@i-am-sijia this is the fix for the intermediate error we discussed yesterday.

i-am-sijia:
Thank you. This fixes the NoneType error, but Sharrow is still crashing at the `non_mandatory_tour_scheduling.vectorize_tour_scheduling.tour_6.interaction_sample_simulate.eval_interaction_utils`, due to the difference in utilities. See attached logs.


[activitysim_nonetype_error.log](https://github.com/ActivitySim/activitysim/files/14827622/activitysim_nonetype_error.log)
[activitysim.log](https://github.com/ActivitySim/activitysim/files/14827691/activitysim.log)

You mentioned Sharrow can find and print out the utility expression(s) that have the biggest value difference. It will be helpful in this case. What setting should I use?

I did not get this error when running the full-scale benchmarking run with Sharrow. So this might just be happening with the small test example. Should I open a separate issue?

i-am-sijia:
Would this PR have any memory implication?

jpn--:
> You mentioned Sharrow can find and print out the utility expression(s) that have the biggest value difference. It will be helpful in this case. What setting should I use?

You just run in debug mode setting a breakpoint here:
https://github.com/ActivitySim/activitysim/blob/2540ede5d6718c20dbd65143205e81c9aabaaf1d/activitysim/core/interaction_simulate.py#L582
and the variable `look_for_problems_here` should point the way.

jpn--:
> Would this PR have any memory implication?

There are no implications for memory usage unless sharrow and non-sharrow disagree on what the utility values are.  In that case, sharrow does gobble up a bunch of memory to write out all the expression values into memory so you can peruse them in the debugger, but this is obviously desirable in this situation.

i-am-sijia:
> > Would this PR have any memory implication?
> 
> There are no implications for memory usage unless sharrow and non-sharrow disagree on what the utility values are. In that case, sharrow does gobble up a bunch of memory to write out all the expression values into memory so you can peruse them in the debugger, but this is obviously desirable in this situation.

That makes sense. I believe the utility comparison only happens in the sharrow compile mode.  This change should not impact if the user is running sharrow production mode, which is when we are usually concerned about memory.

i-am-sijia (approved):

PR #833, opened 2024-03-18
Automatically drop unneeded columns in choosers table

i-am-sijia:
This PR addresses https://github.com/ActivitySim/activitysim/issues/792

- [x] Drop unused columns automatically based on UECs
- [x] Pick up variables used/defined in source code, e.g. custom chooser
- [x] Skip dropping when running household debug tracing
- [x] Skip dropping when running estimation mode
- [x] Pass all existing CI tests
- [x] Run benchmarking runs
- [x] Clean up codes, consolidate dup codes into util funtions

PR #849, opened 2024-04-16
Parking Location Choice Memory Reduction

dhensle:
This PR removes unused columns in the parking location choice model when creating the interaction dataframe combining alternatives and choosers. 

Built on top of the function drop_unused_chooser_columns that is part of #833.


dhensle:
Changes included in #833

PR #870, opened 2024-05-09
Explicit chunking on all interaction simulate models

dhensle:
Adds the option of explicit chunking to all interaction simulate models that were not already hooked-up. These include destination choice, location choice, and scheduling.

Also implemented a feature where the explicit_chunk setting can be less than 1. If less than one, it specifies the fraction.  So `explicit_chunk: 0.1` would mean that there would be 10 chunks.  If greater than 1, explicit_chunk remains the total number of rows in the chooser table.

### DIFF
--- a/activitysim/core/chunk.py
+++ b/activitysim/core/chunk.py
@@ -1213,7 +1213,7 @@ def adaptive_chunked_choosers(
     chunk_tag: str = None,
     *,
     chunk_size: int | None = None,
-    explicit_chunk_size: int = 0,
+    explicit_chunk_size: float = 0,
 ):
     # generator to iterate over choosers
 
@@ -1232,12 +1232,16 @@ def adaptive_chunked_choosers(
 
     chunk_tag = chunk_tag or trace_label
 
+    num_choosers = len(choosers.index)
+
     if state.settings.chunk_training_mode == MODE_EXPLICIT:
-        chunk_size = explicit_chunk_size
+        if explicit_chunk_size < 1:
+            chunk_size = math.ceil(num_choosers * explicit_chunk_size)
+        else:
+            chunk_size = int(explicit_chunk_size)
     elif chunk_size is None:
         chunk_size = state.settings.chunk_size
 
-    num_choosers = len(choosers.index)
     assert num_choosers > 0
     assert chunk_size >= 0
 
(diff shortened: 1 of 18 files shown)


dhensle:
Sharing some testing results for posterity.

Used TransLink's model at a 10% sample size.  No chunking looks like this: 
translink_10pct_no_chunk

I then set chunk_training to explicit with the following explicit_chunk settings for submodels:
* workplace_location: 0.5
* mandatory_tour_scheduling: 0.25
* non_mandatory_tour_destination: 0.5
* non_mandatory_tour_scheduling: 0.5
* trip_destination: 0.5

(notice the y-axis scale difference from the above plot)
translink_10pct_exp_chunk

Run time for no chunking was 113.9 minutes and for explicit chunking was 115 minutes -- very minimal increase in runtime.

activitysim/core/chunk.py
```diff
@@ -1232,12 +1232,19 @@ def adaptive_chunked_choosers(
 
     chunk_tag = chunk_tag or trace_label
 
+    num_choosers = len(choosers.index)
+
+    explicit_and_odd_num_choosers = False
```
jpn--: The value of `explicit_and_odd_num_choosers` is not needed.  It's perfectly fine for some multiple of `rows_per_chunk` to overrun the end of the choosers by a bit, slicing beyond the end of the range.  If it were needed, checking for odd wouldn't be enough, we'd need to adjust based on the inverse of the number of chunks (e.g. 0.25 won't line up unless the total is divisible by 4 not 2)

activitysim/core/chunk.py
```diff
@@ -1260,6 +1267,14 @@ def adaptive_chunked_choosers(
     while offset < num_choosers:
         i += 1
 
+        if (
+            explicit_and_odd_num_choosers
+            & (i == estimated_number_of_chunks)
+            & (rows_per_chunk > 1)
+        ):
+            # last chunk may be smaller than chunk_size due to rounding error
```
jpn--: We don't need to update the rows_per_chunk here, as noted above we can overrun the end of the choosers and be fine.

jpn-- (changes requested):
This looks great.  Couple minor changes to simplify.

activitysim/core/chunk.py
dhensle: Please take a look at the changes in this commit: https://github.com/dhensle/activitysim/commit/1ed94c5f1169a2deb52e8673dc3ba169f431ea8a

I was hitting the assert statement for the alts, (not the choosers) which prompted me to try to adjust the `rows_per_chunk`.  But good point about odd not being good enough.  I removed this functionality and replace the assert statement with a simple check on overflow of the alt index. I think the solution in the above commit fixes the issue (it runs successfully), but suggest you take a look. Thanks!

activitysim/core/chunk.py
dhensle: see above response.

jpn-- (approved):

PR #871, opened 2024-05-14
`protect_columns` for all simulations (choosers, alts, simple simulate, interaction simulate, etc)

i-am-sijia:
This PR includes the following fixes:	

- [x] Add `protect_columns` to all core simulate functions. `protect_columns` is a setting under `compute_settings` that can be used to protect specific columns from being dropped. But it was not fully implemented for all core simulate functions.
- [x] Protecting special columns in the ALT table from being dropped was done previously by hardcoding them in source code, like in simulate_sample.py: https://github.com/ActivitySim/activitysim/blob/1cb48c7acc2c377f4822e2521a19503842aaa495/activitysim/core/interaction_sample.py#L268 This is ok (and efficient) for `tdd` since it’s a generic variable, but less so for `origin_destination` since it’s a variable only relevant to the SANDAG xborder model. Furthermore, this crashes the ARC trip scheduling choice model. This PR moves the non-generic columns from source code to implementation-specific settings.
- [x] Set up `comupte_settings` in trip_scheduling_choice.py
- [x] Add `protect_columns` in parking_location_choice.py

### DIFF
--- a/activitysim/examples/prototype_sandag_xborder/configs/tour_od_choice.yaml
+++ b/activitysim/examples/prototype_sandag_xborder/configs/tour_od_choice.yaml
@@ -55,3 +55,7 @@ SEGMENTS:
 
 LOGSUM_SETTINGS: tour_mode_choice
 LOGSUM_PREPROCESSOR: preprocessor
+
+compute_settings:
+  protect_columns:
+  - origin_destination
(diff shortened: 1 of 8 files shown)


activitysim/examples/prototype_sandag_xborder/configs/tour_od_choice.yaml
```diff
@@ -55,3 +55,7 @@ SEGMENTS:
 
 LOGSUM_SETTINGS: tour_mode_choice
 LOGSUM_PREPROCESSOR: preprocessor
+
+compute_settings:
+  protect_columns:
+  - origin_destination
```
dhensle, line 61: This PR will break the sandag model that's not in the test system... are we keeping track of these breaking changes anywhere?  I think there have been a few of these since we started the Phase 8 work and I don't want them to get lost...

dhensle (approved):

PR #878, opened 2024-07-16
Performance docs

jpn--:
Documentation updates to explain how users can tune ActivitySim models for decent performance.


docs/users-guide/performance/chunking.md
```diff
@@ -0,0 +1,65 @@
...
+in each chunk explicitly, either as an integer number of choosers per chunk, or
+as a fraction of the overall number of choosers. Although the total amount of
+memory engaged for processing any particular chunk is ignored and there is no
+effort to find a "optimal" chunk size, this system is easier to use
+and understand than dynamic chunking, and in practice has been found to be more
+robust and reliable. It requires no "training" and is activated by setting the
+`chunk_training_mode` configuration setting to `explicit`.
+
+This method for chunking does rely upon model developers to have identified the
+memory-hungry components and to have set reasonable explicit chunk sizes for them.
```
dhensle, line 62: Need some description here about how to determine what model run size is -- run with a 10% or so sample single processed with memory profiling on and chunking off, then calculate what the memory needs are.  (Perhaps we should include a little script in the `utilities`portion of the repo to plot the memory profile itself?)

docs/users-guide/performance/chunking.md
```diff
@@ -0,0 +1,65 @@
...
+a problem in interaction-type models, where a large pool of choosers is faced
+with a large set of alternatives.
+
+ActivitySim includes the ability to "chunk" these model components into more
+manageable sized groups of choosers, which can be processed one chunk at a time.
+There is a small overhead associated with chunking, but if the total number of
+chunks is relatively small, the overhead is usually outweighed by the benefits
+in reduced memory usage.
+
+Chunking can be used in two ways in ActivitySim: dynamic and explicit.
```
dhensle: Suggest we recommend explicit chunking here.

docs/users-guide/performance/multithreading.md
```diff
@@ -0,0 +1,102 @@
...
+Some parts of ActivitySim are also multithreaded, using various different
+multithreading technologies. This includes large matrix multiplication operations
+used in certain utility calculations (multithreaded by the MKL library where
+available) as well as computations accelerated using the sharrow library (which
+uses numba to parallelize certain operations), as well as potentially other
+parallelization subroutines in other libraries upon which ActivitySim depends.
+
+In general, it is desirable to fully or partially *disable* multithreading in
+ActivitySim when using multiprocessing, as the two technologies can interfere
+with each other.
```
dhensle, line 12: We should state here that multithreading is off by default when multiprocessing is true.

docs/users-guide/performance/multithreading.md
```diff
@@ -0,0 +1,102 @@
...
+- `OMP_NUM_THREADS` - This environment variable controls the number of threads
+  used by the OpenMP library, which is used by some libraries that ActivitySim
+  depends on.  Setting this to 1 will disable multithreading in OpenMP.
+- `NUMEXPR_NUM_THREADS` - This environment variable controls the number of threads
+  used by the numexpr library, which is used by pandas for certain operations.
+  Setting this to 1 will disable multithreading in numexpr.
+
+The downside of this is that the exact commands needed are platform and
+shell-specific (i.e. the commands for Powershell, Windows `cmd.exe`, and Linux's
+bash are all different).
```
dhensle, line 52: We should specify what the default value of these are... did we end up defaulting them to 1?

dhensle (changes requested):
I think we need a recommended checklist or something similar for people that are handed a model and data and want to run it performantly on a new server.  This would probably be something like: 

1. run sharrow compile with a small hh sample size and save the sharrow cache to be used in subsequent runs
2. run memory profiling with a 10% or so sample to determine memory requirements
3. configure chunking if necessary
4. Run full sample in multiprocessing with 10 cores
5. Experiment with number of cores and threads to find the best performance.

docs/users-guide/performance/index.md
```diff
@@ -0,0 +1,37 @@
...
+These techniques are the focus of this section.
+
+```{tip}
+If you are unsure about the best settings for runtime performance of your model,
+try running with
+
+    multiprocessing: true
+    num_processes: 10
+    chunk_training_mode: explicit
+    sharrow: true
```
i-am-sijia: Would `sharrow: true` fall back to the no-sharrow mode even when the error encountered is not `pandas.eval` related?  This block of settings requires the user to have already run the sharrow compile first in single process, otherwise it will have the file access error when compiling with multiprocess.

docs/users-guide/performance/index.md
```diff
@@ -0,0 +1,37 @@
...
+and to tune that performance to the specific hardware on which the model is being run.
+These techniques are the focus of this section.
+
+```{tip}
+If you are unsure about the best settings for runtime performance of your model,
+try running with
+
+    multiprocessing: true
+    num_processes: 10
+    chunk_training_mode: explicit
```
i-am-sijia: And reference link to Explicit Chunking documentation? Because if the user does not set the `explicit_chunk` in selected components, explicit chunking will not run.

docs/users-guide/performance/index.md
```diff
@@ -0,0 +1,37 @@
...
+of the model components. However, for any given fully developed ActivitySim model,
+there are several techniques that users can employ to improve runtime performance,
+and to tune that performance to the specific hardware on which the model is being run.
+These techniques are the focus of this section.
+
+```{tip}
+If you are unsure about the best settings for runtime performance of your model,
+try running with
+
+    multiprocessing: true
```
i-am-sijia:
```suggestion
    multiprocess: true
```

docs/users-guide/performance/sharrow.md
```diff
@@ -0,0 +1,39 @@
+# Sharrow
+
+Significant performance improvements can be achieved by using the `sharrow` library,
+although doing so requires certain limitations on model design, particularly
+relating to what expressions are allowed in utility specifications.  Generally
+any model specification can be accommodated in `sharrow` by re-writing problematic
+expressions in a more `sharrow`-friendly way, or by moving them to a pre-processor.
```
i-am-sijia, line 7: We should add a bit guidance here on how to find "sharrow-problematic expressions", which is to set `sharrow: test` in the top level setting. If the expression is incompatible with sharrow, or is incorrect in no-sharrow and gets evaluated correctly by sharrow, the user will get a utility not aligned error when running in this mode. We should recommend setting `sharrow: test` if sharrow is run for the first time for an ActivitySim model.

i-am-sijia (changes requested):
In addition to David's comments. I'd like to add:

- We did a lot of memory and run time profiling in phase 9A. It would be nice to add an `Example Model Performance` section in the user docs to document the example 1-zone and 2-zone model performance.
- Recommend setting `sharrow: test` to catch components that are sharrow incompatible if a region is running sharrow for the first time with their model. It will be helpful to give some guidance on how to find the problematic expressions. It be helpful to also document some example incompatible expressions and how to fix them, e.g., move them into preprocessor.

docs/users-guide/performance/chunking.md
jpn--: Yes, done.

docs/users-guide/performance/multithreading.md
jpn--: Description of default threading settings added.

docs/users-guide/performance/multithreading.md
jpn--: Added language about defaults.

docs/users-guide/performance/index.md
jpn--: Added note to this effect.

docs/users-guide/performance/sharrow.md
jpn--: I have added some discussion on this to the developer's guide page on sharrow, which is linked from the user's guide.    I think the "user" docs should assume the model configs are sharrow-safe and just talk about turning sharrow on or off; making a sharrow-safe model is a "developer" level project.

docs/users-guide/performance/index.md
jpn--: Added more detail in the "steps" section that clarifies to have compiled first.

docs/users-guide/performance/chunking.md
jpn--: @dhensle That's a great idea to include a script in the repo to plot the memory profile.  Can you create it?

jpn--:
> * We did a lot of memory and run time profiling in phase 9A. It would be nice to add an `Example Model Performance` section in the user docs to document the example 1-zone and 2-zone model performance.

I will open another PR with a summary of performance results and a description of the new two "canonical" examples.

docs/users-guide/performance/chunking.md
dhensle: Just made a commit to this branch and added the little jupyter notebook I have been using to plot the memory profile.  I put it in other_resources/scripts.

jpn--:
I believe I have addressed all the review comments.  @dhensle @i-am-sijia

dhensle (approved):

i-am-sijia (approved):
Looks great!

PR #882, opened 2024-07-26
Release notes v1.3

jpn--:
Release notes for v1.3, inserted into documentation.


jpn--:
@dhensle @i-am-sijia did I missing anything important?

docs/dev-guide/changes.md
```diff
@@ -0,0 +1,231 @@
...
+```yaml
+chunk_training_mode: explicit
+```
+
+Then, for model components that may stress the memory limits of the machine,
+the user can specify the number of choosers in each chunk explicitly, either as an integer
+number of choosers per chunk, or as a fraction of the overall number of choosers.
+This is done by setting the `explicit_chunk` configuration setting in the model
+component's settings.  If the `explicit_chunk` value is 0 or missing, then no
+chunking is applied for that component.  The `explicit_chunk` values are ignored
```
dhensle: Suggest we say here that if  explicit_chunk > 1 then it corresponds to the number of rows. If fraction, then it corresponds to the fraction of the total number of choosers.

dhensle (commented):
Looks great! Do we also want to mention the creation of the ABM3 and MTC example model repos?  I expect people will want to be aware of and use them.

docs/dev-guide/changes.md
```diff
@@ -0,0 +1,231 @@
...
+### Automatic conversion of string data to categorical
+
+Version 1.3 introduces a new feature that automatically converts string data
+to categorical data.  This reduces memory usage and speeds up processing for
+large models.  The conversion is done automatically for string columns
+in most chooser tables.
+
+See code updates in [PR #782](https://github.com/ActivitySim/activitysim/pull/782)
+and [PR #863](https://github.com/ActivitySim/activitysim/pull/863)
+
```
i-am-sijia, line 188: Suggest adding here that - There is also an optional downcasting of numeric data available if the user want to further reduce memory. But we suggest using it with caution and it is defaulted to `False` (not downcasting numerics) because: 1) changing the precision of numeric data could cause results to change slightly and impact a previous calibrated model result, and 2) downcasting to lower byte data types, e.g., int8, can cause numeric overflow in downstream components if the numeric variable is used in math calculations such as squaring, that would result in values beyond the int8 limit. We recommend users to review their model specification and review the result change if they choose to downcast numeric data. To downcast, users can set the `downcast_int` and `downcast_float` to `True` in the top level model configuration file (typically `settings.yaml`)

docs/dev-guide/changes.md
```diff
@@ -0,0 +1,231 @@
...
+significant reduction in memory usage for large models.
+
+Dropping columns may also cause problems if the model is not correctly configured.
+If it is desired to use this feature, but some required columns are being dropped
+incorrectly, the user can specify columns that should not be dropped by setting the
+[`protect_columns`](activitysim.core.configuration.base.ComputeSettings.protect_columns)
+setting under [`compute_settings`](activitysim.core.configuration.base.ComputeSettings).
+This allows the user to specify columns that should not be dropped, even if they are
+not apparently used in the model component.
+
```
i-am-sijia, line 158: Example: https://github.com/ActivitySim/activitysim/blob/67820ad32789f59217608b5311e9a2a322d029ed/activitysim/examples/prototype_sandag_xborder/configs/tour_od_choice.yaml#L59-L61

i-am-sijia (changes requested):
Looks good to me. Suggested two edits.

Issue #874, opened 2024-06-05
Sharrow debug in simulate.py not working


i-am-sijia:
**Describe the bug**
This came up when I ran ABM3 100% sample with sharrow set to `test`. See issue https://github.com/ActivitySim/sandag-abm3-example/issues/18. I found a work around to debug but it would be better if we can fix sharrow...

In sharrow `test` mode, when the utility returned by sharrow does not match the non-sharrow utility, sharrow attempts to retrieve data and print out `possible problematic expressions` to help debugging. See code here:
https://github.com/ActivitySim/activitysim/blob/29d12bcc054d9206b01318b352d1c33f12c7dca0/activitysim/core/simulate.py#L800-L809
However, currently the run will get stuck (no crash no response) in line 807 `_sh_flow_load = sh_flow.load(sh_tree)`. Not sure if this is broken due to some recent changes.

Also, for many example models, the sharrow `test` mode has only been run with small samples, and issues like this seem to get discovered especially when we run large samples (100% HH). This means when agencies adopt sharrow, we should probably recommend running `test` mode with a large (enough) sample.

**To Reproduce**

- ActivitySim: d98f776af559038e1588c7f07d58df7151965cb3
- Sharrow: https://github.com/ActivitySim/sharrow/commit/5b968ae86a632ea2751ce46f251f83e5f00d05d8
- ABM3: https://github.com/ActivitySim/sandag-abm3-example/commit/f0fbc8fb9e2020b76dbe5decff405c8a169e134c
- Set sharrow to `test` and run ABM3.

**Expected behavior**
It would be nice to have line 807 working and return the problematic expressions.

i-am-sijia:
Update: I ran the [small ABM3 test example](https://github.com/ActivitySim/sandag-abm3-example/blob/f0fbc8fb9e2020b76dbe5decff405c8a169e134c/test/test_sandag_abm3.py#L181) with sharrow on, the model got stuck (no crash, no response, no termination) in the same location. I believe this is why our GitHub Actions test has been failing with error code 143: https://github.com/ActivitySim/activitysim/actions/runs/9209634995/job/25335037344

bhargavasana:
@i-am-sijia can this be closed?

PR #884, opened 2024-08-05
Bugfix for trip scheduling choice

jpn--:
The trip scheduling model currently was set with `allow_zero_probs=True` but immediately after running the choice model it includes an assertion that all the choices are valid.  Really we want to disallow zero probabilities.  By setting this to false we also automatically engage underflow protection on the probabilities when the utility dtype is float32 (as happens with sharrow enabled).

Also this PR temporarily removes a problematic block of code when tracing and sharrow are both enabled.


i-am-sijia (approved):
@jpn-- I commented out another section in simulate.py that also calls the sh_flow.load() that would cause the model to hang. It's in this pr: https://github.com/camsys/activitysim/pull/15

i-am-sijia:
> @jpn-- I commented out another section in simulate.py that also calls the sh_flow.load() that would cause the model to hang. It's in this pr: [camsys#15](https://github.com/camsys/activitysim/pull/15)

It was raised in this issue: https://github.com/ActivitySim/activitysim/issues/874

Issue #890, opened 2024-09-20
Sharrow cache not being written to directory specified in sharrow_cache_dir


JoeJimFlood:
**Describe the bug**
I'm currently trying to upgrade ABM3 to work with the latest version of ActivitySim. While testing sandag-abm3-example I noticed that the Sharrow cache was being written to a location in the C:\Users folder for my account on the modeling server I'm running it on. When I tried setting the [sharrow_cache_dir](https://github.com/ActivitySim/activitysim/blob/main/activitysim/core/configuration/filesystem.py#L386) setting, it still looked for/wrote to that same location in C:\Users. I was hoping that we could store the cache in a centralized location and copy from it or even point to it when our modelers run the model. We have 12 modelers and 15 modeling servers so modelers needing to compile Sharrow every time a modeler runs the model on a server they haven't run it on yet isn't feasible.

**To Reproduce**
1. Set `sharrow_cache_dir` in the settings file for a run.
2. After the skims are read in and once some steps are run, open the logfile.
3. Look for the text "flow.cache_dir" and look to where the cache is being written.

**Expected behavior**
I was expecting the Sharrow cache to be written to the directory specified by `sharrow_cache_dir`.

**Screenshots**
![SharrowCacheDir](https://github.com/user-attachments/assets/93e365e0-f2ab-4902-8788-966061eef997)

PR #893, opened 2024-10-01
Sharrow Cache Dir Setting

dhensle:
Fix for #890. Also fixes the same issue for `profile_dir` and `pipeline_file_name` settings.

### DIFF
--- a/activitysim/core/workflow/state.py
+++ b/activitysim/core/workflow/state.py
@@ -478,14 +478,11 @@ class State:
             include_stack=False,
         )
 
-        # the settings can redefine the cache directories.
-        cache_dir = raw_settings.pop("cache_dir", None)
-        if cache_dir:
-            if self.filesystem.cache_dir != cache_dir:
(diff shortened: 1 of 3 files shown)


activitysim/core/workflow/state.py
```diff
@@ -478,15 +478,12 @@ def load_settings(self) -> State:
             include_stack=False,
         )
 
-        # the settings can redefine the cache directories.
-        cache_dir = raw_settings.pop("cache_dir", None)
```
jpn--, line 482: We used "pop" here, this takes this setting off the `raw_settings`.  The revision leaves the setting in place... should we be popping all these values?

I think we need to write a test [or tests] to make sure this is solving the problem well.

jpn--:
This appears to work.  We need to have a test to confirm that, but writing one is some effort and in the meantime consortium members will benefit from having this fixed.  I opened #902 to record that we want this tested, and will merge this for now.

Issue #897, opened 2024-10-02
Location choice logsum overwritten with mode choice logsum in estimation mode


dhensle:
**Describe the bug**
The location choice logsum gets overwritten by the mode choice logsum in location choice when running in estimation mode.

**To Reproduce**
Run the estimation example with the DEST_CHOICE_LOGSUM_COLUMN_NAME set in the workplace or school location yaml.

**Expected behavior**
The location choice logsum should be the location choice logsum, not the mode choice logsum.

**Screenshots**
workplace location and modechoice logsums are the same, as seen in the below trace
![image](https://github.com/user-attachments/assets/91fb6eb6-5726-4a58-82b5-2e90d33b49b8)

**Additional context**
This only occurs in estimation mode. The previous code unnecessarily overwrote the location choice logsum with the mode choice logsum of the selected alternative.  Based on the code, it seems this was done because the previous coder did not fully comprehend the difference between the location and modechoice logsums, thus treating them as the same.  This was also likely complicated by the fact that we now allow the user to return both the modechoice and location choice logsums separately.

dhensle:
Handled in #917

dhensle:
This came back! Not sure why...

PR #898, opened 2024-10-02
Location and modechoice logsum written out in estimation mode

dhensle:
Fix for #897 

Additional small change that sets the probability of added choices (required if sampling and the override choice is not in the sample) to 0 so the results are not affected by the added alternative.


dhensle:
subsumed by #917

PR #932, opened 2025-03-18
Update to use pandas v2.*

jpn--:
Addresses #794.

The update from pandas 1.x to 2.x introduces a number of small but material changes that affect ActivitySim:

- DataFrame `Index` objects are all one class with different datatypes, instead of being different classes (e.g. there is no more `Int64Index` class).
- The `read_csv` function by default now interprets "None" as a missing value (i.e. NaN) instead of being the Python object `None`.
- The `groupby` operation, when applied to categorical data, now sorts the categories in the result unless told not to (resulting in different order of rows in outputs for some operations).
- A simple `df.join()` also potentially sorts the resulting rows differently unless an explicit `sort` argument is given.
- `Index` objects no longer can be checked as `is_monotonic` but instead need `is_monotonic_increasing`.
- The handling of dtypes appears to have improved in some instances, where dtypes used to be promoted by some operations now they are not (e.g. variables that are originally int16 used to become int64 after some operations and now they don't).

This pull request includes several changes across multiple files to address these pandas changes. The most important changes include modifications to sorting operations, error handling in logging, and the introduction of a new `fast_eval` function to optimize DataFrame evaluations, because the regular `pandas.eval` has some significant performance degradations.

### Data Handling Improvements:
* [`activitysim/abm/models/disaggregate_accessibility.py`](diffhunk://#diff-9e3d1fb7696ae6786a4d224ecd062ac8d527d8ac4809d6183cba2309c81e03e6L578-R578): Added sorting operations to ensure data consistency in `expand_template_zones`, `create_proto_pop`, and `merge_persons` methods. [[1]](diffhunk://#diff-9e3d1fb7696ae6786a4d224ecd062ac8d527d8ac4809d6183cba2309c81e03e6L578-R578) [[2]](diffhunk://#diff-9e3d1fb7696ae6786a4d224ecd062ac8d527d8ac4809d6183cba2309c81e03e6L651-R653) [[3]](diffhunk://#diff-9e3d1fb7696ae6786a4d224ecd062ac8d527d8ac4809d6183cba2309c81e03e6R729)
* [`activitysim/abm/models/school_escorting.py`](diffhunk://#diff-1d67ac6f064a19a419b67fc8c36b56a9a1a2f18cc4dc2ed0588876fc9b1ea713L637-R637): Reset index for `escort_bundles` to maintain data integrity.
* [`activitysim/abm/models/util/school_escort_tours_trips.py`](diffhunk://#diff-f34bad9be0028fa549ffc474aa51f96c0b6b31a429a284955e8b7b87555eaa6eL356-R356): Reset index for `create_chauf_escort_trips` and `create_escortee_trips` methods to ensure proper data handling. [[1]](diffhunk://#diff-f34bad9be0028fa549ffc474aa51f96c0b6b31a429a284955e8b7b87555eaa6eL356-R356) [[2]](diffhunk://#diff-f34bad9be0028fa549ffc474aa51f96c0b6b31a429a284955e8b7b87555eaa6eL542-R542)

### Error Handling Enhancements:
* [`activitysim/abm/models/input_checker.py`](diffhunk://#diff-19dbdcbe489ba2503d590f90042dc3eb0f114b4fa723f263b5b8ac41557eca26R304-R312): Added try-except blocks to improve error logging for dataframe and element-wise validators. [[1]](diffhunk://#diff-19dbdcbe489ba2503d590f90042dc3eb0f114b4fa723f263b5b8ac41557eca26R304-R312) [[2]](diffhunk://#diff-19dbdcbe489ba2503d590f90042dc3eb0f114b4fa723f263b5b8ac41557eca26L328-R339)
* [`activitysim/core/test/_tools.py`](diffhunk://#diff-07c277af029f4af6f1a47c70f1286602c8f476c970bacbd64c8f14997db42167L168-R169): Enhanced error reporting in `progressive_checkpoint_test` by including exception details.

### Evaluation Process Optimization:
* Introduced `fast_eval` function in `activitysim/core/fast_eval.py` to optimize DataFrame evaluations by handling special characters in column names and improving performance.
* Updated references to `df.eval` in `activitysim/core/interaction_simulate.py` and `activitysim/core/simulate.py` to use `fast_eval` for better performance and consistency. [[1]](diffhunk://#diff-1121dd7397f0098ca85161fa1246a6ac29103aa5f66e893c8bf63af56c6ed46aR17) [[2]](diffhunk://#diff-1121dd7397f0098ca85161fa1246a6ac29103aa5f66e893c8bf63af56c6ed46aL290-R291) [[3]](diffhunk://#diff-1121dd7397f0098ca85161fa1246a6ac29103aa5f66e893c8bf63af56c6ed46aL559-R560) [[4]](diffhunk://#diff-3b8c25f307a55dd3bdd60b73a025893bc6174504ed67c7ae2702bdd3eac51193L659-R660) [[5]](diffhunk://#diff-3b8c25f307a55dd3bdd60b73a025893bc6174504ed67c7ae2702bdd3eac51193L916-R917)

### Miscellaneous Changes:
* [`activitysim/cli/create.py`](diffhunk://#diff-5859dfac8b422de4721c9d22bf928b683218303abe3b73d5bf595afbbbc7b51fR5): Replaced `pkg_resources` with `importlib.resources` for resource handling to modernize the codebase. [[1]](diffhunk://#diff-5859dfac8b422de4721c9d22bf928b683218303abe3b73d5bf595afbbbc7b51fR5) [[2]](diffhunk://#diff-5859dfac8b422de4721c9d22bf928b683218303abe3b73d5bf595afbbbc7b51fL24-R32) [[3]](diffhunk://#diff-5859dfac8b422de4721c9d22bf928b683218303abe3b73d5bf595afbbbc7b51fL180-R183)
* [`activitysim/core/los.py`](diffhunk://#diff-7c1c031529711ac262c8a56155be3b10a61a3f906b6c1ca38f5efcd927a5e499L783-R791): Improved handling of data types to prevent overflow in `get_mazpairs` method.


jpn--:
The changes I have made in this new branch have greatly improved runtime performance while using pandas 2.x.  

non-sharrow test timings for pandas 1.x:
```
58.60s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_mp
53.71s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc
53.66s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_chunkless
53.23s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_recode
```
first attempt non-sharrow test timings for pandas 2.x (#838):
```
148.50s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc
148.14s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_chunkless
147.83s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_recode
140.09s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_mp
```
revised non-sharrow test timings for pandas 2.x (this PR, #932):
```
65.06s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_mp
58.10s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_chunkless
58.10s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc
57.38s call     activitysim/examples/prototype_mtc/test/test_mtc.py::test_mtc_recode
```

We can see that there is still a modest runtime cost to using pandas 2.x, on the order of 10% slower, but nowhere near the cost of the first attempt, which was ~200% slower.  Achieving no runtime penalty appears to be possible, but it would require accessing non-public pandas functions which might break in the future, see [here](https://github.com/driftlesslabs/activitysim/blob/501e2499d7dfaab0dd9cf17e2cfc4449a57a59c6/activitysim/core/fast_eval.py#L34-L37)

Note all of these runtime issues are exclusively non-sharrow, as sharrow evaluation completely bypasses the `pandas.eval` function that is the source of our problem.

activitysim/core/fast_eval.py
```diff
@@ -0,0 +1,98 @@
...
+    # CHANGED FROM PANDAS: do not call df.dtype inside the dict comprehension loop
+    # This update has been made in https://github.com/pandas-dev/pandas/pull/59573,
+    # but appears not to have been released yet as of pandas 2.2.3
+    dtypes = df.dtypes
+
+    return {
+        clean_column_name(k): Series(
+            v, copy=False, index=df.index, name=k, dtype=dtypes[k]
+        ).__finalize__(df)
+        for k, v in zip(df.columns, df._iter_column_arrays())
```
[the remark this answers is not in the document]
jpn--: Thanks, Copilot, for reiterating back my comment that I put in this code 10 lines above.  By default, our new code *does* use the public API. This alternative code path through a private method is left available here in case its speed gains are worth activating.

JoeJimFlood:
Have any runtime comparisons been done with the full sandag-abm3-example or anything larger than the 25-zone prototype_mtc example? I'm concerned about nonlinearity in the relationship between the size of a model and the runtime.

i-am-sijia:
> Have any runtime comparisons been done with the full sandag-abm3-example or anything larger than the 25-zone prototype_mtc example? I'm concerned about nonlinearity in the relationship between the size of a model and the runtime.

Hi @JoeJimFlood, that is a valid concern. As I am reviewing this PR, I can perform the run time test with the full size example SANDAG.

i-am-sijia:
I ran the full size sandag-abm3-example with this PR (e.g., pandas 2.x) and the main branch (e.g., pandas 1.4) and would like to share some quick initial reports. I do have some other comments which I will post separately.

For both runs, I used:

- Sharrow: False
- multiprocess: True
- num_processes: 5
- explicit_chunk: 0.2 (for select components)

The run time is almost the same for the two runs, see below. Pandas 2.x runs faster for some components but not the others, e.g., it's faster in non-mandatory tour scheduling but slower in mandatory tour scheduling, which could be just runtime noise. In total, pandas 2.x took ~5 mins longer which is probably negligible. The total run time is comparable to the run time I reported during Phase 9A: https://github.com/ActivitySim/sandag-abm3-example/issues/9#issuecomment-2236832539. This PR does shorten the run time for pandas 2.0 as it promised.

I have not checked if the results of the two run are the same, I will check that.

image

jpn--:
I see long sequences where one version is like ~10% faster, or ~10% slower, in sequential contiguous blocks across fairly disparate component types.  This strongly suggests much of the runtime differences are external noise from other subprocesses or other issues (e.g. the server got too hot and throttled the compute for a couple minutes).

JoeJimFlood:
Thanks for running that @i-am-sijia! The 1.5% increase in using Pandas v2 vs v1 is encouraging to see.

activitysim/core/interaction_simulate.py
```diff
@@ -287,7 +288,7 @@ def to_series(x):
                         if expr.startswith("@"):
                             v = to_series(eval(expr[1:], globals(), locals_d))
                         else:
-                            v = df.eval(expr, resolvers=[locals_d])
+                            v = fast_eval(df, expr, resolvers=[locals_d])
```
i-am-sijia, line 291: When future pandas releases address the regular pandas.eval performance issue, ActivitySim would not need its own `fast_eval()`, and can continue using the regular eval. Anticipating this, should we add a condition here for pandas version = 2.2.x?

conda-environments/activitysim-dev-base.yml
```diff
@@ -31,7 +31,6 @@ dependencies:
 - ipykernel  # so this env will appear in jupyter as a selection
 - isort
 - jupyterlab
-- larch = 5.7.*
```
i-am-sijia: Why is larch removed?

.github/workflows/core_tests.yml
```diff
@@ -275,11 +275,11 @@ jobs:
           - region: Standard 1-Zone Example (MTC)
             region-org: ActivitySim
             region-repo: activitysim-prototype-mtc
-            region-branch: extended
+            region-branch: pandas2
           - region: Standard 2-Zone Example (SANDAG)
             region-org: ActivitySim
             region-repo: sandag-abm3-example
-            region-branch: main
+            region-branch: pandas2
```
i-am-sijia, line 282: The changes in the example repos updated the expected result parquet files. I suppose those are just index object type, data type changes?

activitysim/core/fast_eval.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    from pandas import Series
+    from pandas.core.computation.parsing import clean_column_name
+
+    if isinstance(df, pd.Series):
+        return {clean_column_name(df.name): df}
+
+    # CHANGED FROM PANDAS: do not even convert the arrays to pd.Series, just
+    # give the raw arrays to the compute engine. This is potentially a breaking
+    # change if any of the operations in the eval string require a pd.Series.
```
i-am-sijia, line 32: This means we cannot have a utility expression specific to pd.Series, such as .map(). Could we create a check on user specified expressions to alert/avoid this? Could this be done as part of https://github.com/ActivitySim/activitysim/pull/936? And the alternative is to move strings require pd.Series to preprocessors?

i-am-sijia (changes requested):
I'm curious about the implications on dependence lock and expression rules. With this PR, ActivitySim will use its own `fast_eval()` and some rewrite version of internal pandas methods until pandas releases an official version (say pandas 3.0) that fixes our problem at hand. Is the plan for us to be locked with pandas 2.2 and fast_eval.py until then? Otherwise we are adding an overhead to maintain the compatibility of fast_eval() when we'd use pandas >2.2.  In terms of expression rules, I saw the comment related to pd.Series in fast_eval.py, was wondering if we should proactively alert users about that.

activitysim/core/interaction_simulate.py
jpn--: I would prefer not to pin to versions of pandas within the code itself.  I think the better strategy is to leave the code as-is, and pin to "pandas>=2.2,<3.0" in the dependencies.  I expect the forthcoming dependency locking work will implement that kind of thing anyway.  Then, if in the future a pandas update causes fast_eval to break, we fix it then.  I think it's more likely that the fast_eval will still work and just not be necessary, but there's no telling when that will be.

conda-environments/activitysim-dev-base.yml
jpn--: Oops!  This was an accident.  (Larch 5.7 from conga-forge is no longer compatible with Apple silicon, so I need to turn it off for my personal development.  This will be fixed soon when we pin ActivitySim to larch>=6.0)

.github/workflows/core_tests.yml
jpn--: They are not exclusively index object type changes, there are also changes in the sort order of some operations.  ActivitySim imposes sorting pretty much everywhere that sort order matters, but in a handful of cases it doesn't matter, and/or or there are default dataframe index values that are ignored by ActivitySim but are technically still present in intermediate data frames that are being regression tested, and which appear as re-ordered with pandas 2.x

activitysim/core/fast_eval.py
jpn--: Proactively catching every utility expression specific to pd.Series is tricky, as we'd need to parse the expression and analyze it, and we'd need something like a comprehensive list of code features that work with pd.Series but not a simple array.

The solution for now is down on line 87, in the try-except:  first we try to eval the string assuming it will work with just arrays, and if it fails we retry with the make-it-series solution.  This relies on the assumption that eval expressions that depend on the inputs being pd.Series will raise an exception when run on arrays, and not return values that are different from expected.  In my experience the exception-raising is much more likely, but the latter is perhaps possible. I have found and fixed similar problems when comparing legacy-vs-sharrow code -- which is a slightly different kind of problem but in a very similar vein, and extremely challenging to diagnose. Only a very careful and rigorous testing of models will ever uncover this kind of thing.

activitysim/core/interaction_simulate.py
i-am-sijia: Thanks. I'm ok with not to pin versions within the code itself then.

Regarding the following, thinking about leaving some breadcrumbs for our future selves, something like "if the run crashes at this line, check your pandas version and potentially switch to the default pandas eval".

> Then, if in the future a pandas update causes fast_eval to break, we fix it then.

activitysim/core/fast_eval.py
i-am-sijia: Thank you for the thoughtful answer. Yea, failing silently sounds like a tough challenge. I don't know if I can come up with a better solution than what you've implemented. Perhaps before the try-except block, scan the expression string for know pd.Series features (acknowledge it won't be a comprehensive list) and force it go to the make-it-series solution, so that it reduces the reliance on the exception?

jpn--: I think parsing these expressions trying to catch these rare problems is going to add an undesirable amount of runtime overhead to every component, on every model run.  On the flip side, I think it's a better solution to expect model designers will test their specifications and check they are running correctly.  One useful tool to do this testing is by running with sharrow in test mode, which completely circumvents this problem.

activitysim/core/interaction_simulate.py
jpn--: I've added this note to the doctoring for the `fast_eval` function.

i-am-sijia (approved):
Thank you for responding to my comments. I'll approve this PR.

PR #948, opened 2025-06-02
Pandas 2 `assign_in_place()` fix with categoricals

i-am-sijia:
ActivitySim models often call the [`activitysim.core.util.assign_in_place(df, df2, ...)`](https://github.com/ActivitySim/activitysim/blob/4063cd6b5213327697fe0d4e85961ebbab61b5a3/activitysim/core/util.py#L317) method to update existing values in `df` using values from `df2`, or to add new columns from `df2`. When performing the former, it calls [`df.update(df2)`](https://github.com/ActivitySim/activitysim/blob/4063cd6b5213327697fe0d4e85961ebbab61b5a3/activitysim/core/util.py#L343) to update common columns in place. 

In Pandas 2.x, `df.update(df2)` will raise a `TypeError` if the common column `C` in `df` is of categorical dtype and `df2['C']` contains category value(s) that are not present in `df['C']`'s categories.

This issue was encountered by SANDAG in the trip preprocessor of their airport model, see discussion #946 . None of the existing ActivitySim example models exhibit this issue.

The solution in this PR first makes sure `df2['C']` is a categorical type, and then unions the two categoricals before calling df.update(df2).

### DIFF
--- a/activitysim/core/util.py
+++ b/activitysim/core/util.py
@@ -354,7 +354,7 @@ def assign_in_place(df, df2, downcast_int=False, downcast_float=False):
                 # when df and df2 column are both categorical, union categories
                 from pandas.api.types import union_categoricals
 
-                uc = union_categoricals([df[c], df2[c]])
+                uc = union_categoricals([df[c], df2[c]], sort_categories=True)
                 df[c] = pd.Categorical(df[c], categories=uc.categories)
                 df2[c] = pd.Categorical(df2[c], categories=uc.categories)
 
@@ -444,7 +444,7 @@ def auto_opt_pd_dtypes(
                 else:
                     df[col] = pd.to_numeric(df[col], downcast="float")
         # Skip if the column is already categorical
-        if pd.api.types.is_categorical_dtype(dtype):
+        if isinstance(dtype, pd.CategoricalDtype):
             continue
         # Handle integer types
         if pd.api.types.is_integer_dtype(dtype):
--- a/activitysim/core/util.py
+++ b/activitysim/core/util.py
@@ -340,6 +340,23 @@ def assign_in_place(df, df2, downcast_int=False, downcast_float=False):
     common_columns = df2.columns.intersection(df.columns)
     if len(common_columns) > 0:
         old_dtypes = [df[c].dtype for c in common_columns]
+        # in pandas 2.x, update a categorical column with any new categories will cause TypeError
+        # so we need to add the new categories first
+        # this is a workaround for pandas 2.x, see discussion in
+        # https://github.com/ActivitySim/activitysim/discussions/946
+        for c in common_columns:
+            if isinstance(df[c].dtype, pd.CategoricalDtype):
+                if not isinstance(df2[c].dtype, pd.CategoricalDtype):
+                    # if df column is categorical, but df2 column is not
+                    # convert df2 column to categorical then union categories
+                    df2[c] = df2[c].astype("category")
+
+                # when df and df2 column are both categorical, union categories
+                from pandas.api.types import union_categoricals
+                uc = union_categoricals([df[c], df2[c]])
+                df[c] = pd.Categorical(df[c], categories=uc.categories)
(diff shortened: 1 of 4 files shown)


activitysim/core/util.py
```diff
@@ -340,6 +340,24 @@ def assign_in_place(df, df2, downcast_int=False, downcast_float=False):
...
+            if isinstance(df[c].dtype, pd.CategoricalDtype):
+                if not isinstance(df2[c].dtype, pd.CategoricalDtype):
+                    # if df column is categorical, but df2 column is not
+                    # convert df2 column to categorical then union categories
+                    df2[c] = df2[c].astype("category")
+
+                # when df and df2 column are both categorical, union categories
+                from pandas.api.types import union_categoricals
+
+                uc = union_categoricals([df[c], df2[c]])
```
jpn--: I recommend setting `sort_categories=True` here, which will help make sure that categories populate in a stable order under multiprocessing, which will help with reproducibility/stability, especially when multiprocessing.

jpn--:
Has this been tested?  Can we write a test for it?  It looks like it should work, but it also may have unexpected consequences (e.g. from reordering categories).  I see the current tests are passing, but they were also passing without this so it's clear our current tests are not exercising the problem being solved.

jpn--:
@i-am-sijia, I just also saw your comment in https://github.com/ActivitySim/activitysim/discussions/946#discussioncomment-13338181 .  Can we check if bringing "sort_categories=True" to the other existing usage[s] of union_categoricals will fail any of our existing tests?

i-am-sijia:
@jpn-- I added `sort_categories=True` to all `union_categoricals()`. But now I am having second thoughts. Most categorical variables are created with choice alternatives being the pre-defined categories, using the same sort sequence of how the alternatives are defined. Categoricals may not be in the alphabetically order, but they are sorted and include all possible values. This makes categoricals stable under multiprocessing.
https://github.com/ActivitySim/activitysim/blob/ba4a12e85b3e4e794124bf39d4c2d006063462b1/activitysim/abm/models/mandatory_tour_frequency.py#L137-L141
We need `union_categoricals()` when new category values are added. You are right that under multiprocessing, the sequence of how the new categories are being appended may be unstable. But if we set `sort_categories=True` during union, it will also change the original sequence of the old category values.

1. Should we sort all categoricals alphabetically at creation, not just when union is happening? Otherwise you are changing the sequence when union. So that they will be consistent through out.
2. With `sort_categories=True`, whenever a new category is append, it re-sorts everything instead of appending, which also impacts the stability. Would this be a problem for Sharrow, i.e., triggers re-compilation?
3. Side note - In general, I think we do not want to set `ordered=True`, unless we want to write expressions directly compare categories. I reverted the ordered mode categories. [[1](https://github.com/ActivitySim/activitysim/pull/948/files#diff-37ea4cb55f0720bcea7c0ccb904c797254d6d4c74379b6d537ccfe17c851f03eR91-R92)] _(I don't recall why I specifically made it ordered, I left a note in there in case it causes us trouble in the future)_

In terms of unit tests. I was thinking we can start with the following:
- Concatenating two categorical columns with different categories
- Overwriting a categorical column with a non-categorical column
- Overwriting a categorical column with a categorical column with different categories

jpn--:
Maybe we are getting too far ahead of ourselves.

I agree that most categoricals are fully enumerated at creation time, in some logical order that may be meaningful to modelers (e.g. driving, then transit, then nonmotorized).  And, most categoricals in ActivitySim are logically unordered.

For sharrow, any change in a categorical data type (adding categories, removing unused categories, reordering them, making them "ordered" when they were previously just nominal), literally any change at all will result in recompiling.  As long as we get stable sort ordering most of the time, we can survive with corner cases that don't have stable order (e.g. the sample is too small and some unusual tour type doesn't always get included).

So, maybe we just merge this to fix the bug encountered as reported at the top, and leave a more complete solution to efficiently handling categoricals until ActivitySim 2.x?

Issue #965, opened 2025-08-12
Explicit Park-and-Ride modeling and capacity constraint


dhensle:
**Is your feature request related to a problem? Please describe.**
Park-and-ride is currently onerous to model.  For most ActivitySim implementations, the selection of the actual PnR lot used for park-and-ride trips is handled through hyper-pathing in the commerical assignment software packages.  These packages will generate probabilities across multiple paths including potentially different PnR lots and distributes the PnR demand coming out of ActivitySim along those paths.  

Skims are then generated for PnR based on these paths.  The skims are very expensive -- need to generate them for both directions (since the auto portion vs tansit portion changes based on inbound vs outbound) and all the normal transit variables like in vehicle time, cost, wait time, etc.   Additionally, ActivitySim does not have any mechanism to implement capacity constraints on the PnR lots.

**Proposed PnR methodology**
We want to implement an explicit park-and-ride lot choice model in ActivitySim.  This would run before tour mode choice.  The choosers for this model are all tours that have walk-transit access at their destination end. The alternatives are all lots with non-zero PnR capacity.

Utilities used to make the choice are broken down into two components - auto from tour origin to PnR lot and walk-transit from PnR lot to destination.  The number of parking spaces would act as the size term. 

The PnR lot choice model output is the zone_id of the PnR lot the tour would use if that tour were to choose park-and-ride in the tour mode choice model.  PnR utilities in tour mode choice would be built similarly as the lot choice model -- auto costs to lot + transit costs to destination.

Capacities at each lot can then be calculated after tour mode choice.  Tours that arrive after the lot is full (or sampled randomly) can be re-simulated with a new PnR lot choice option with full lots removed and updated tour mode choice utilities.  PnR lot choice and tour mode choice would iterate until all PnR lots are under capacity. This is very similar to how the simulation-based constraint mechanism in workplace location works.

Some minor updates to the write trip matrices is needed to split the PnR trips into auto and transit portions for the output demand matrices.

**Implications**

- Allows the removal of costly PnR skims
- Allows the ability to build utilities directly to PnR lots and eventually estimate the coefficients for PnR lot choice
- Allows the ability to track the PnR lot capacity and have tour mode choice respond
- PnR lot choice also needs to be run when calculating logsums in upstream models (there is an option to skip this but then the logsum would not include PnR availability)

**Limitations / Future Enhancements**

- Iterating only with tour mode choice does not allow people to change other aspects of their tour.
[...]

PR #1001, opened 2025-10-08
Park-and-ride lot choice and capacity

dhensle:
Code changes for #965


jpn--:
@copilot resolve the merge conflicts in this pull request

activitysim/abm/models/util/park_and_ride_capacity.py
```diff
@@ -0,0 +1,502 @@
...
+    One buffer for adding up the number of park-and-ride lot choice zone ids to calculate capacities.
+    One other buffer keeping track of the choices for each tour so we can choose which ones to resimulate.
+    This function is called before the multiprocesses are kicked off in activitysim/core/mp_tasks.py
+    """
+
+    # get landuse and person tables to determine the size of the buffers
+    persons = state.get_dataframe("persons")
+
+    # don't know a-priori how many park-and-ride tours there are at the start of the model run
+    # giving the buffer a size equal to the number of persons should be sufficient
```
jpn--, line 473: This is *probably* fine most of the time, but in the outside chance that the number of PnR tours does exceed the number of persons, the model is going to crash hard later, probably just after line 159 where `pad` will become negative.

activitysim/abm/models/util/park_and_ride_capacity.py
```diff
@@ -0,0 +1,502 @@
...
+                synced_choices = pd.concat(
+                    [mp_choices_df, choices[["pnr_zone_id", "start"]]],
+                    axis=0,
+                    ignore_index=False,
+                )
+                # sort by index (tour_id)
+                synced_choices = synced_choices.sort_index()
+
+                # now append any additional rows need to get size back to original length
+                pad = len(self.shared_pnr_choice) - len(synced_choices)
```
jpn--, line 156: We should intercept here if `pad` is negative, maybe crash gracefully explaining that the number of PnR tours was unexpectedly greater than the number of persons, instead of crashing with an arcane "negative dimensions are not allowed" error from deep in this code.  (see initialization comment below)

activitysim/abm/models/util/park_and_ride_capacity.py
```diff
@@ -0,0 +1,502 @@
...
+
+            # count the total number of pnr choices being resimulated
+            pnr_counts = (
+                over_capacitated_tours.pnr_zone_id.value_counts()
+                .reindex(self.shared_pnr_occupancy_df.index)
+                .fillna(0)
+                .astype(int)
+            )
+
+        elif self.model_settings.RESAMPLE_STRATEGY == "random":
```
jpn--, line 351: There's no testing of `RESAMPLE_STRATEGY == "random"`

activitysim/abm/models/util/park_and_ride_capacity.py
```diff
@@ -0,0 +1,502 @@
...
+            # using ActivitySim's RNG to make choices for repeatability
+            current_sample, rands = logit.make_choices(state, probs)
+            current_sample = current_sample[current_sample == 1]
+
+            # filtering choosers to only those tours selected for resimulation in this subprocess
+            choosers = choosers[choosers.index.isin(current_sample.index)]
+
+            # count the total number of pnr choices being resimulated
+            pnr_counts = (
+                current_sample.pnr_zone_id.value_counts()
```
jpn--: I tried making a simple test for "random" but it failed here, `current_sample` seems to be a pd.Series.

The correct code might be:
```
pnr_counts = (
    tours_in_cap_zones.loc[current_sample.index, "pnr_zone_id"]
    .value_counts()
    .reindex(self.shared_pnr_occupancy_df.index)
    .fillna(0)
    .astype(int)
)
```

activitysim/abm/models/util/park_and_ride_capacity.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def set_choices(self, choices):
+        """
+        aggregate individual location choices to modeled_size by zone and segment
+
+        Parameters
+        ----------
+        choices : pandas.Series
+            zone id of location choice indexed by person_id
+        segment_ids : pandas.Series
```
jpn--: This function does not take `segment_ids`

jpn-- (changes requested):
Mostly good.  That several agencies have collaborated and tested this already gives some extra confidence.   Only a couple minor bugs/concerns to address.

activitysim/abm/models/util/park_and_ride_capacity.py
dhensle: Yes, good point.  Put a check on this and it should now fail gracefully with an informative error message.

activitysim/abm/models/util/park_and_ride_capacity.py
dhensle: agreed. Put a check on `pad`. See below comment.

activitysim/abm/models/util/park_and_ride_capacity.py
dhensle: Fixed the issue with this resample strategy and added a unit test.

activitysim/abm/models/util/park_and_ride_capacity.py
dhensle: fixed!

activitysim/abm/models/util/park_and_ride_capacity.py
dhensle: removed

activitysim/core/steps/output.py
```diff
@@ -205,7 +205,7 @@ def write_data_dictionary(state: workflow.State) -> None:
     if txt_format:
         with open(state.get_output_file_path(txt_format), "w") as output_file:
             # get max schema column widths from omnibus table
-            col_width = {c: schema_df[c].str.len().max() + 2 for c in schema_df}
+            col_width = {c: int(schema_df[c].str.len().max()) + 2 for c in schema_df}
```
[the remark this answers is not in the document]
dhensle: accepted recommendation

activitysim/abm/models/park_and_ride_lot_choice.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def return_no_choices(state, choosers: pd.DataFrame, original_index=None) -> pd.Series:
+    logger.debug(
+        "No choosers with transit accessible destinations found. Returning -1 as park-and-ride lot choice."
+    )
+    # need to drop rng channel that we created before trn_accessible_choosers
+    state.get_rn_generator().drop_channel("pnr_lot_choice")
+    index = choosers.index if original_index is None else original_index
+    return pd.Series(data=-1, index=index)
```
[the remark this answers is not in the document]
dhensle: accepted recommendation (removed "Only" in the comment string though)

activitysim/abm/models/atwork_subtour_mode_choice.py
```diff
@@ -64,38 +66,33 @@ def atwork_subtour_mode_choice(
...
+                network_los=network_los,
+                model_settings=None,
+                choosers_dest_col_name="destination",
+                choosers_origin_col_name="workplace_zone_id",
+                estimator=None,
+                pnr_capacity_cls=None,
+                trace_label=tracing.extend_trace_label(trace_label, "pnr_lot_choice"),
+            )
+        else:
+            subtours_merged["pnr_zone_id"].fillna(-1, inplace=True)
```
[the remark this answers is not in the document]
dhensle: Cleaned this up in an if-else statement instead of the weird nested if thing I originally had.  I didn't accept this quite as recommended -- I don't want to add pnr_zone_id to the output table if that functionality as a whole is not turned on.

activitysim/abm/models/park_and_ride_lot_choice.py
```diff
@@ -0,0 +1,413 @@
...
+    land_use: pd.DataFrame,
+    network_los: los.Network_LOS,
+    model_settings: ParkAndRideLotChoiceSettings | None = None,
+    choosers_dest_col_name: str = "destination",
+    choosers_origin_col_name: str = "home_zone_id",
+    estimator=None,
+    model_settings_file_name: str = "park_and_ride_lot_choice.yaml",
+    pnr_capacity_cls: ParkAndRideCapacity | None = None,
+    trace_label: str = "park_and_ride_lot_choice",
+) -> None:
```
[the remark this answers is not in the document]
dhensle: fixed

activitysim/abm/models/park_and_ride_lot_choice.py
```diff
@@ -0,0 +1,413 @@
...
+
+    Provides another entry point for the model which is useful for getting pnr locations while creating logsums.
+    """
+
+    if model_settings is None:
+        model_settings = ParkAndRideLotChoiceSettings.read_settings_file(
+            state.filesystem,
+            "park_and_ride_lot_choice.yaml",
+        )
+
```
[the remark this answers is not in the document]
dhensle: fixed

activitysim/abm/models/util/park_and_ride_capacity.py
```diff
@@ -0,0 +1,502 @@
...
+                synced_choices = synced_choices.sort_index()
+
+                # now append any additional rows need to get size back to original length
+                pad = len(self.shared_pnr_choice) - len(synced_choices)
+                new_arr_values = np.concatenate(
+                    [
+                        synced_choices["pnr_zone_id"].to_numpy(np.int64),
+                        np.zeros(pad, dtype=np.int64),
+                    ]
+                )
```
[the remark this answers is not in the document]
dhensle: This was addressed in the upstream comments.

activitysim/abm/tables/disaggregate_accessibility.py
```diff
@@ -16,7 +16,7 @@
 
 
 def find_nearest_accessibility_zone(
-    state: workflow.State, choosers, accessibility_df, method="skims"
+    state: workflow.State, model_settings, choosers, accessibility_df, method="skims"
 ):
```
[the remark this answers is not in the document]
dhensle: Removed changes in this file. Not sure why this is showing up as a change.  model_settings is not used, so just removed that from the function and the one function call location.

dhensle:
@jpn-- I have address all the comments and merged in the latest main code.  I think this should be ready to merge (assuming all the tests pass as expected)

jpn--:
A re-review found a few more issues.

1. **Resampled PnR lot choices are not written back to `tours`.**

   `final_choices` is updated before `pnr_zone_id` is attached to `choices_i`; subsequent lot choices modify only the merged `choosers` frame. At completion, only `final_choices` is assigned back to `tours`. Thus a tour’s mode can be calculated using a newly selected lot while downstream outputs retain its original lot. That undermines both capacity results and the new PnR-leg matrices. 

2. **PnR-enabled mandatory-tour scheduling logsums use the wrong destination column.**

   `_compute_logsums` runs lot choice with hard-coded `choosers_dest_col_name="destination"` before resolving `DESTINATION_FOR_TOUR_PURPOSE`. Mandatory scheduling normally uses `workplace_zone_id` or `school_zone_id`, so enabling `include_pnr_for_logsums` will access a missing/wrong column. 

3. **`RESAMPLE_STRATEGY: random` remains incorrect.**

   Under multiprocessing, `choices_synced` and `probs` contain tours from every process, but each subprocess RNG channel contains only its local tour IDs. `random_for_df(probs)` therefore tries to locate foreign IDs in local RNG state and raises `KeyError`.  Even single-process sampling uses `occupancy / capacity - 1`. On the initial pass, the rate needed to remove the expected excess is `(occupancy - capacity) / occupancy`. For occupancy 3 and capacity 2, the code uses 0.5 rather than 1/3; for 4 and 2 it resamples everyone rather than half.

I pushed a commit that includes tests demonstrating all three problems.  Then I pushed a commit with fixes.  @dhensle if you approve I will merge this.

dhensle:
These are good catches and the fixes look good.  Thanks @jpn--. I approve and think this is ready for merge!

jpn-- (approved):

bhargavasana:
I would like to record this comment for future PRs. I think it would be great to include at least a basic https://github.com/ActivitySim/activitysim/tree/main/docs update for a major feature like this. SANDAG is also interested in this feature, but I am not sure what exactly has been implemented and how exactly to use it. Maybe agencies already using this feature can contribute for the benefit of other members.
cc @bwentl @bettinardi @stefancoe @joecastiglione

Issue #408, opened 2021-03-30
Clean up command prompt logging


bstabler:
| Idea          | Level-of-effort | Notes |  Priority | 
| ------------- | ------------- |   ------------- | ---- |
| Clean up command prompt logging | Hours depending on LOE | Be more intentional about what is logged to the console vs. log file since there's a lot of info currently. | Medium |

PR #1004, opened 2025-10-20
Clean up command prompt logging

JoeJimFlood:
This pull request cleans up the logging that was raised in #408, primarily by downgrading multiple instances of `logger.info()` to `logger.debug()`. Guidance on what logging level to use for each logging message was also added to the logging documentation. Statements were targeted for demotion by running the prototype_mtc example with the logging level set to `INFO` and prioritizing statements that were repetitive and not helpful for simply determining where the model was at (but could be useful for debugging). The changes in this pull request result in the size of the output logfile (and by proxy number of messages printed in the console) for the prototype_mtc example decreasing from 3290 KB to 139 KB, a 95.8% reduction.


jpn-- (approved):
I have reviewed this and found it works as described.  There are a few stray messages we could still get and remove from the main logs but that need not hold up approving this.

PR #1007, opened 2025-10-28
Expand UV installation and usage instructions

### LINKED ISSUE #995: Develop and test robust, multi-platform UV-based installation for v1.4.x and 1.5.x
Phase 10 introduced UV-based packaging and installation of ActivitySim to enhance version control and reproducibility, and to ensure smooth installation of ActivitySim environments across different platforms and configurations. ActivitySim’s online user [documentation](https://activitysim.github.io/activitysim/develop/users-guide/modelsetup.html#installing-activitysim) has been revised with updated instructions for both users and developers on installing and running ActivitySim using UV. Integration tests within GitHub Actions have also been updated to use UV instead of conda for installing and testing ActivitySim example features and models. 

The purpose of this task is to identify and implement updates to the existing UV instructions and documentation by testing these instructions across the range of partner agency ActivitySim IT and computing environments.  This effort will include verifying dependency resolution, handling version conflicts, and evaluating performance across different user workflows.  Improvements to the UV instructions will be identified by implementing UV within agencies using ActivitySim, revealing gaps in the current installation instructions. The updated instructions will include recommended installation procedures, troubleshooting steps, and environment management strategies to support consistent and reliable use of the ActivitySim software.

i-am-sijia:
This pull request substantially revises and expands the documentation for installing and setting up ActivitySim using the UV package and project manager. The changes clarify recommended installation methods for different user types, provide detailed step-by-step instructions for installing UV and ActivitySim (both from PyPI and from source with a lockfile), and introduce a comprehensive Q&A section to address common user scenarios and troubleshooting.

**Expanded and Detailed UV Installation Instructions**
* Added step-by-step instructions for installing UV on Windows (including user and global installs), updating PATH, and verifying installation, with links to official resources.

**Comprehensive ActivitySim Installation Options Guidance and User Recommendations**
* Provided thorough instructions for installing ActivitySim using UV both from PyPI and from source with a lockfile, including explanations of the implications for dependency management and reproducibility.
* Clarified which installation method is recommended for different users, specifying criteria such as familiarity with Python, need to customize code, and management of virtual environments.
* Added a comparison table to help users choose between installation options based on their needs (official release, development, reproducibility, etc.).

**New Q&A Section for Common User Scenarios**
* Introduced a Q&A section addressing how to add additional packages, manage conflicting dependencies, share working environments, handle dependency issues, and clarify when to activate virtual environments versus using `uv run`.

### DIFF
--- a/docs/users-guide/modelsetup.rst
+++ b/docs/users-guide/modelsetup.rst
@@ -240,15 +240,15 @@ instructions can be found `here <https://git-scm.com/downloads>`_.)
   git clone https://github.com/ActivitySim/activitysim.git  
   cd activitysim
 
-Run the `uv sync` command to create a virtual environment using the lockfile. It will initialize a virtual environment within the `activitysim` directory
+Run the `uv sync --locked` command to create a virtual environment using the lockfile. It will initialize a virtual environment within the `activitysim` directory
 and install ActivitySim and all its dependencies exactly as specified in the `uv.lock` file. 
 The virtual environment is a hidden folder within the current directory called 
 `.venv` and operates the same way as Python's classic *venv*.
 
 ::
 
-  uv sync
-  # or uv sync --no-editable
+  uv sync --locked
+  # or uv sync --locked --no-editable
 
 It is worth pointing out that by default, *uv* installs projects in 
 editable mode, such that changes to the source code are immediately reflected 
@@ -344,10 +344,10 @@ How to share my working environment with others?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 You can share your working environment with others by sharing the `uv.lock` file located in your project directory. 
 This file contains the exact versions of all packages and dependencies used in your project. 
-Others can recreate the same environment by running the `uv sync` command in a new project directory containing the shared `uv.lock` file.
+Others can recreate the same environment by running the `uv sync --locked` command in a new project directory containing the shared `uv.lock` file.
 ::
   # Copy uv.lock file to new project directory
-  uv sync
+  uv sync --locked
 
 If I use the From PyPI option to install ActivitySim, would I run into dependency issues?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/users-guide/modelsetup.rst
+++ b/docs/users-guide/modelsetup.rst
@@ -92,8 +92,9 @@ There are two recommended ways to install ActivitySim:
 
 2. Using the :ref:`UV Package and Project Manager`
 
-The first is recommended for users who do not need to change the Python code and are on Windows, 
-and the second is recommended for users who need to change/customize the Python code.
+The first is recommended for users who are new to Python and use Windows, do not actively create and manage Python virtual environments, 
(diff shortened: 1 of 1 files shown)


jpn-- (approved):

jpn--:
Do we want to consider recommending locking *uv* on a script file for model application use?  An [example](https://github.com/driftlesslabs/sandag-abm3-example/blob/sync_production_v15.3.1/scripts/run-large-sharrow.py) is here.  I'm not sure if this is easier or harder for an agency application.

i-am-sijia:
@jpn--  my understanding is that `uv lock --script one_single_script.py` is a minimal set up, is fast, but is limited to single scripts (what happens if agencies have multiple scripts?). Is there any additional benefit compared to just setting up a UV project/environment in the normal way? The locking process is not fundamentally different than setting up a UV project/environment so I am open to be convinced either way. Some online posts suggest: "for model applications, use script locking for exploration/prototyping, transition to full project when moving to production or when complexity grows."

docs/users-guide/modelsetup.rst
```diff
@@ -125,173 +126,273 @@ UV Package and Project Manager
...
+
+  # Copy .python-version file to new project directory (if exists)
+  copy path\to\shared\.python-version .
+  # Copy pyproject.toml file to new project directory
+  copy path\to\shared\pyproject.toml .
+  # Copy uv.lock file to new project directory
+  copy path\to\shared\uv.lock .
+
+  # Recreate the same environment
+  uv sync --locked
```
dhensle, line 393: Suggest we add an additional Question here: "Can other users on the same server or machine use my already created virtual environment?"
My understanding is the answer is: "No.  When UV installs the dependencies, paths are created specific to the user.  As such, if another user tries to activate your virtual environment, that environment will not work for them."

docs/users-guide/modelsetup.rst
```diff
@@ -125,173 +126,273 @@ UV Package and Project Manager
...
+Many agencies use commercial software that have Python APIs and dependencies that may conflict with ActivitySim dependencies. 
+In such cases, we also recommend creating a separate *uv* project for the commercial software and managing them independently from ActivitySim.
+
+::
+
+  # Open Command Prompt
+  mkdir emme_project
+  cd emme_project
+  uv init --python 2.7
+  # Then copy the emme.pth file (provides EMME API handshakes) from the Emme installation directory to emme_project/.venv/Lib/site-packages/
```
dhensle, line 359: Suggest we explicitly state here that users can easily switch between different UV environments by just activating the one they want to use. (Do you need to deactivate your current environment first?)

docs/users-guide/modelsetup.rst
i-am-sijia: The answer is Yes, but it requires admin rights and additional setup. I added it to the Q&A and provided some instructions.  SANDAG was able to do this. I haven't tested it myself because I do not have admin rights on WSP machines. I think we should advise "This is doable but we do not recommend it. We recommend sharing the lockfile and uv sync"

docs/users-guide/modelsetup.rst
i-am-sijia: added.

docs/users-guide/modelsetup.rst
jpn--: I think what @i-am-sijia wrote in this version of the docs is fine.  If I were writing it I'd write it in an even more strongly worded manner; an ideal use case is a fully containerized environment where ActivitySim runs in a way where no other user and no other process can interfere with it.

Issue #951, opened 2025-06-11
PNUM person variable hard-coded in ActivitySim


dhensle:
**Describe the bug**
Joint tour participants and the joint tour frequency and composition models look for a PNUM attribute in the persons table.  If you do not have PNUM in your person's file, the model will crash. 

**To Reproduce**
Run an ActivitySim model using the joint_tour_participants or joint_tour_frequency_composition models without PNUM in the input tables.

**Expected behavior**
There should not be a dependency on having PNUM in the input persons table.  The user should be allow to name it something different (e.g. person_number, p_num, etc.) or omit it completely.

**The Fix**
joint_tour_frequency_composition model looks for PNUM == 1 to define a point person for the joint tour.  Instead, we can just use the first person on the joint tour.

joint_tour_participants model uses PNUM to create a unique joint tour participant_id.  Fix is to just create a PNUM column on the fly if none exist.

jpn--:
Closed by #953, and also an identical bug closed in #1008

PR #1008, opened 2025-10-30
fix(estimation): correct the participant_num assignment logic for estimation mode in joint tour participation 

asiripanich:
This PR fixes two related issues in the joint-tour code paths:

## `joint_tour_participation.py` (8f13299)
**Problem**: In estimation mode, the code could re-assign participant_num rather than using the original survey participant_num, which could change the identity of the tour leader (ActivitySim treats participant_num == 1 as the tour leader). This broke association of the joint tour with the correct person.

**Fix**: When running in estimation mode, we now merge participant_num from the survey's joint_tour_participants table by participant_id so the original participant numbering (and the tour leader) is preserved.

## `joint_tour_frequency.py` (2e7629e6)
**Problem**: The code assumed a PNUM column existed; where PNUM was missing, code could fail or behave unpredictably.

**Fix**: If PNUM is missing, the code now selects the first person in the household to act as the PNUM (keeps behavior deterministic and prevents failures).

cc: @dhensle

### DIFF
--- a/activitysim/abm/models/joint_tour_frequency.py
+++ b/activitysim/abm/models/joint_tour_frequency.py
@@ -137,7 +137,15 @@ def joint_tour_frequency(
     # - but we don't know the tour participants yet
     # - so we arbitrarily choose the first person in the household
     # - to be point person for the purpose of generating an index and setting origin
-    temp_point_persons = persons.loc[persons.PNUM == 1]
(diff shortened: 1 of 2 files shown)


asiripanich:
Related to #951 .

jpn-- (approved):

Issue #1016, opened 2025-11-12
Simulate Chooser Columns and Logsum Chooser Columns Settings Not needed


dhensle:
**Is your feature request related to a problem? Please describe.**
The settings SIMULATE_CHOOSER_COLUMNS and LOGSUM_CHOOSER_COLUMNS were added as memory hacks to reduce the size of the interaction dataframe for models that combine chooser tables with alternative tables.

We have since added a utility that will automatically drop all columns not needed from the table before this merge is made based on the spec (util.drop_unnecessary_columns()).

As such, these settings are now unnecessary and serve as "gotchas" in the code.

**Describe the solution you'd like**
Lets deprecate these settings!

Issue #1015, opened 2025-11-12
Global constants not available in all models


dhensle:
**Describe the bug**
Global constants are not available in every model specification.

**To Reproduce**
1. put constant in constants.yaml
2. try to use that constant in the following models: location choice, destination choice, and tour scheduling.  It will not be recognized

**Expected behavior**
We should always have access to our global constants

**Additional context**
Useful for things like applying a factor to convert, say, km to miles using a single constant throughout the modeling system.

dhensle:
Related to #1079

PR #1017, opened 2025-11-12
Access to global constants and removing chooser col filtering

dhensle:
Fixes for #1016 and #1015.  (These are in the same PR since they both arose when cleaning a single model's configs up.)


activitysim/abm/models/school_escorting.py
```diff
@@ -369,6 +368,21 @@ class SchoolEscortSettings(BaseLogitComponentSettings, extra="forbid"):
         Multinomial logit model.
     """
 
+    @property
```
jpn--: Changing this to a property is the backwards fix: it keeps it readable but not writable.  But we need it to be writable (i.e. the values can appear in the YAML file and it will be allowed) but there should be a field validator that tells you about the deprecation.

jpn-- (changes requested):

activitysim/abm/models/school_escorting.py
dhensle: Yes, saw the tests were failing cause of this.  Was initially just copying the deprecation warnings in trip destination, not thinking hard enough about the differences.. anyways, updated them to, I believe, the functionality that we are expecting here.

jpn--:
@dhensle I have review this, and found there are substantive failures in the tests for this PR.  Some of them look like issues with columns getting dropped at the wrong time, so that there are duplicate values when there shouldn't be.  But the placeholder_multiple_zone test is getting a different problem: the model runs without crash or failure, but some of the math comes out different.   I spent a few minutes trying to find the issue but it's somewhere deep.  Can you investigate?

jpn--:
Closing in favor of #1094 and #1095

Issue #846, opened 2024-04-04
alternatives files have inconsistent formatting


jpn--:
```
     # FIXME setting index as "Alt" causes crash in estimation mode...
     # happens in joint_tour_frequency_composition too!
     # alts = simulate.read_model_alts(state, model_settings.ALTS, set_index="Alt")
     alts = simulate.read_model_alts(state, model_settings.ALTS, set_index=None)
     alts.index = alts["Alt"].values
```

I noticed all other models' alternatives files have the alternative column named and **hardcoded** in source code as `alt`, but for school escorting, it is `Alt`.  It's probably better to keep the column name consistent as `alt`, even though we are not addressing the "hardcoded" part at this moment.

_Originally posted by @i-am-sijia in https://github.com/ActivitySim/activitysim/pull/777#discussion_r1546518616_

I think we should treat this issue separately from this PR.  If we change it here, it will have implications for people who are already using the model (i.e. they would have to change their alternatives file). This is larger than simply changing it here as the Alt name is coded in a couple of places in this file and I would like to keep this PR focused on the BayDAG contributions.  Thus, I think we should track it in a separate and stand-alone issue. Thoughts?

_Originally posted by @dhensle in https://github.com/ActivitySim/activitysim/pull/777#discussion_r1550927471_

i-am-sijia:
@jpn-- Would it be a good idea to have an AlternativeSettings class, that checks alt file?

PR #1018, opened 2025-11-13
Index names consistently set to "alt" for alternatives files

dhensle:
Fix for #846.

In addition to CI tests, also successfully ran the SANDAG (cropped) estimation data through estimation mode and confirmed the creation of school escorting and joint tour frequency and composition EDBs.


jpn-- (approved):

jpn--:
This looks fine to me.  @dhensle  Just one (new) thing to do:  please add a section to `docs/dev-guide/changes.md` about this change, under the `Upcoming Changes` section at the top.

dhensle:
@jpn-- I have update the change log for your review.  Also, I tried to repair the progressive mtc test by unzipping, changing the Alt column name to alt in the school escorting bundle that was failing, and re-zipped it back up, but now I am getting some weird error about not finding the input checker log.  Is there something different you do to create the zip file that maybe I messed up?

jpn--:
@dhensle Typically when I need to regenerate the progressive tests it's because the numbers have changed slightly, so I just delete the regression targets and re-run, which should automatically regenerate the files.  But for a change like this where the numbers are all the same and it's just a header that's changed, what you describe should work -- and I'm surprised it didn't.

docs/dev-guide/changes.md
```diff
@@ -13,6 +13,15 @@ branch (i.e., the main branch on GitHub), but not yet released in a stable versi
 of ActivitySim.  See below under the various version headings for changes in 
 released versions.
 
+## v1.5.2
+
+### Alternative file naming consistency
+
+We have enforced a naming scheme for alterntive config files to have the index name
```
jpn--:
```suggestion
We have enforced a naming scheme for alternative config files to have the index name
```

Issue #899, opened 2024-10-02
Trace files from the same run should all have the same hash


dhensle:
**Is your feature request related to a problem? Please describe.**
Trace files currently written out have a unique hash applied to them.  This allows for multiple runs to be saved and compared.  However, it can be hard to keep track of which trace file goes with which run:
![image](https://github.com/user-attachments/assets/12187079-674e-4bd4-90bc-a43c9e77cd1b)

**Describe the solution you'd like**
All trace files from the same run should use the same hash.

andkay:
@dhensle There already [exists ](https://github.com/camsys/activitysim/blob/e345f9774e14d0eac7fe01b2e22aeca28c2b821f/activitysim/core/workflow/tracing.py#L39) a `RunID` hash that looks like it supposed to be used to [suffix trace files ](https://github.com/camsys/activitysim/blob/e345f9774e14d0eac7fe01b2e22aeca28c2b821f/activitysim/core/workflow/tracing.py#L254). 

Guessing that the issue is that when trying to write trace files, the RunID is either unavailable or the `state.filesystem.get_trace_path` method is using a default time-based hash for some other reason. But let me know if I am barking up the wrong tree here.

dhensle:
@andkay From my read of the existing runID creation code, it looks like the hash is created using the current time.  Does this get created only once at the start of the run?  What happens in mp mode -- will each subprocess have a different time?

I don't think you are barking up the wrong tree, but I am not too familiar with the details of how the current tracing hash was originally implemented.

andkay:
It's definitely hashing the time, but my assumption would certainly be that something called a "RunID" _should_ be created once and be stable across subprocesses. In any case, sounds like I am correct that this is the first port of call to investigate. 

Either the RunID isn't actually uniquely identifying a run based on the launch time -- or there is something going on with the tracing that doesn't feed this information through in which case a different hash based on the current time will be created when the trace file path gets created automatically.

JoeJimFlood:
[...]

PR #1021, opened 2025-11-18
Fix trace IDs so they have the same hash

JoeJimFlood:
This pull request changes the code so that only one Run ID will be generated for the entire model run, and not just for each subprocess. This is being done by adding a module in the core directory called run_id.py that defines a `RunID` object. This is then read in by every `state` object that is generated. As a result, any tracing files for a single run will all have the same hash, which will solve Issue #899.


i-am-sijia (approved):
I've reviewed this pull request. It's working as described.

Issue #892, opened 2024-09-26
Add global option to skip households in simulation when failed


i-am-sijia:
**Is your feature request related to a problem? Please describe.**
----
In production runs, sometimes ActivitySim crashes because one or two choosers cannot make a choice in the simulation. The common error message when this happens is e.g., no choice is available.

As a user of ActivitySim, agency X would like to have the option to skip households if they fail in any components, without crashing the entire ActivitySim run. This is extremely useful in production runs, especially when the agency is under time crunch to get RTP runs done and does not have time / resource to find out why the simulation fails for just a handful of households.

See original comment by @i-am-sijia https://github.com/ActivitySim/activitysim/pull/821#issuecomment-1992029253

**Describe the solution you'd like**
----

This is implemented in CT-RAMP2 models (perhaps also in DaySim?) - When a household is failing in the simulation, skip it and exclude it in the subsequent components and exclude it in the output. Keep a list of such households that failed and write them out in the end, give the user a warning in the end with the number of households that failed. If it's just a small number of households, then it's probably an edge case and user can choose to ignore them and trust the model result. But if it's a large number of households that failed, then it suggests bigger issues like inputs error or specification error, then user can investigate. 

This solution allows the model run to complete, at the same time gives user information on the scale of potential issues behind the scene.

**Describe alternatives you've considered**
----

**Additional context**
----

PR #1023, opened 2025-11-20
Add global option to skip households on simulation failure

i-am-sijia:
This PR addresses issue #892.

**Status:** Draft - core functionality implemented, remaining: logging and documentation. Update: logging and documentation are added.

## Summary
Adds a new `skip_failed_choices` setting that allows ActivitySim to skip households that cause choice model failures instead of being masked and forced a choice by the `overflow_pretection` or crashing the entire simulation. This improves robustness for large-scale production runs where occasional edge cases shouldn't halt the model.

## Key Changes

**Core Logic (`activitysim/core/`)**
- Modified `logit.py` to temporarily mark failed choices with `-99` and track skipped households in state
- Added functionality in `state.py` to dynamically drop data that belong to the skipped household. Further updated the logic to only trigger this when new households are being skipped.
- Updated `simulate.py` and `interaction_sample_simulate.py` to handle failed choices
- Enhanced `report_bad_choices()` to log skipped households
- Added code in `simulate.py` to automatically print out debug tracing for one failed choice per model component, limited to simple simulate models only.

**Configuration**
- Added `skip_failed_choices: bool = True` setting in `top.py`
- Added `fraction_of_failed_choices_allowed = 0.1` setting in `top.py`. Users can optionally specify different values in the settings.yaml to set the threshold of failed/skipped households (relative to sample households) above which the run should terminate.
- Added code in `cli/run.py` to print out a summary of skipped households by model component at the end of the run.

**Model Updates (`activitysim/abm/models/`)**
- Retained `household_id` in chooser columns across location choice models for tracking
- Modified shadow price calculations to exclude skipped households
- Updated trip matrices to adjust trip matrix expansion factor. The adjustment is based on the remaining households' weight vs the initial sampled households' weight.
- Added null trip mode handling in trip mode choice and school escorting

**Testing**
- Added new unit test `test/skip_failed_choices` to deliberately trigger failures and validates households being skipped.
- Added tests with different threshold of skipped households.

**Documentation**
- Added documentation in `docs/users-guide/modelsetup.rst` and `docs/users-guide/ways_to_run.rst`.

## Usage
```yaml
# settings.yaml
skip_failed_choices: true #default
fraction_of_failed_choices_allowed: 0.1 #default
```

Access skipped household info via:
- `state.get("num_skipped_households", 0)`
- `state.get("skipped_household_ids", dict())`

Final count logged at end of simulation in the `activitysim.log`.

## Notes
- Defaults to `True` for production robustness
- Disables overflow protection when active to surface underlying issues
- Temporary code adds `household_id` to chooser columns (remove when PR #1017 merges)


i-am-sijia:
For components where we intentionally allow some choices to fail without being skipped, the `allow_zero_probs` option is already set to `True`, which exempts them from this new feature. Therefore, it may not be necessary to introduce this as a component-level setting. In addition, keeping it as a global setting helps reduce manual errors, avoid confusion, and prevent inconsistencies across the implementation.

activitysim/abm/models/util/school_escort_tours_trips.py
```diff
@@ -1043,6 +1043,10 @@ def force_escortee_trip_modes_to_match_chauffeur(state: workflow.State, trips):
         f"Changed {diff.sum()} trip modes of school escortees to match their chauffeur"
     )
 
+    # trip_mode can be na if the run allows skipping failed choices and the trip mode choice has failed
+    if state.settings.skip_failed_choices:
+        return trips
```
jpn--: Rather than just being OK here, code should:
1. emit a warning that N choices have failed, and
2. check if N exceeds configured failure thresholds (e.g. not more than 1% of all choices), in which case an error should be raised.

activitysim/abm/models/location_choice.py
```diff
@@ -1072,6 +1087,23 @@ def iterate_location_choice(
...
+        # so that their choices are not considered in shadow price calculations
+        # first append household_id to choices_df
+        choices_df = choices_df.merge(
+            persons_merged_df[["household_id"]],
+            left_index=True,
+            right_index=True,
+            how="left",
+        )
+        if len(choices_df) > 0:
+            choices_df = choices_df[
```
jpn--: Should check here how many choices are getting dropped, and
1. emit a warning that N choices have failed, and
2. check if N exceeds configured failure thresholds (e.g. not more than 1% of all households), in which case an error should be raised.

activitysim/abm/models/trip_matrices.py
```diff
@@ -307,6 +307,12 @@ def write_trip_matrices(
                 .TAZ.tolist()
             )
 
+    # print out number of households skipped due to failed choices
+    if state.settings.skip_failed_choices:
+        logger.info(
```
jpn--: logging level for this should be higher than "info"; at least warning if not "error" level.

activitysim/abm/models/trip_matrices.py
```diff
@@ -393,6 +399,21 @@ def write_matrices(
...
         logger.error("Missing MATRICES setting in write_trip_matrices.yaml")
 
+    hh_weight_col = model_settings.HH_EXPANSION_WEIGHT_COL
+    if hh_weight_col:
+        if state.get("num_skipped_households", 0) > 0:
+            logger.info(
+                f"Adjusting household expansion weights in {hh_weight_col} to account for {state.get('num_skipped_households', 0)} skipped households."
+            )
+            # adjust the hh expansion weights to account for skipped households
+            adjustment_factor = state.get_dataframe("households").shape[0] / (
```
jpn--: Should the adjustment to expansion weights be based not on the number of skipped households, but instead on the original expansion weight total and the total expansion weight of the households we dropped?  E.g. if we drop a household with a giant expansion weight that's a bigger thing than dropping a household with a smaller one.

activitysim/abm/models/trip_mode_choice.py
```diff
@@ -382,6 +389,11 @@ def trip_mode_choice(
     # need to update locals_dict to access skims that are the same .shape as trips table
     locals_dict = {}
     locals_dict.update(constants)
+    if state.settings.skip_failed_choices:
+        mask_skipped = trips_merged["household_id"].isin(
```
jpn--: mask_skipped should be the same for trips_df and trips_merged, right?  We only need to calculate the mask once.

activitysim/core/configuration/top.py
```diff
@@ -781,6 +781,13 @@ def _check_store_skims_in_shm(self):
     should catch many common errors early, including missing required configurations or specified coefficient labels without defined values.  
     """
 
+    skip_failed_choices: bool = True
+    """
+    Skip households that cause errors during processing instead of failing the model run.
+
+    .. versionadded:: 1.6
+    """
+
```
jpn--, line 792: Need additional setting[s] to set thresholds for how many skips are OK and when it's too many and should be an error.

activitysim/core/logit.py
```diff
@@ -87,6 +88,27 @@ def report_bad_choices(
...
+                continue
+            if hh_id not in skipped_household_ids:
+                skipped_household_ids.add(hh_id)
+                num_skipped_households += 1
+            else:
+                continue
+        state.set("num_skipped_households", num_skipped_households)
+        state.set("skipped_household_ids", skipped_household_ids)
+
+        logger.debug(
```
jpn--: logger level here should be `warn` not `debug`

jpn-- (changes requested):
Mostly looks fine.  We discussed (possibly while @i-am-sijia was out) that the skipping should not be unlimited; the idea is we want to skip problems when problems are few and probably not seriously consequential.  If the problems are many there's little benefit of allowing a simulation to run to completion.

activitysim/abm/models/util/school_escort_tours_trips.py
```diff
@@ -1043,6 +1043,10 @@ def force_escortee_trip_modes_to_match_chauffeur(state: workflow.State, trips):
         f"Changed {diff.sum()} trip modes of school escortees to match their chauffeur"
     )
 
+    # trip_mode can be na if the run allows skipping failed choices and the trip mode choice has failed
+    if state.settings.skip_failed_choices:
```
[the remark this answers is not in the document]
i-am-sijia: Added warning and additional assertion.

activitysim/core/simulate.py
```diff
@@ -1485,11 +1487,15 @@ def eval_nl(
...
 
+    if state.settings.skip_failed_choices is not None:
+        skip_failed_choices = state.settings.skip_failed_choices
+
     if no_choices.any():
         logit.report_bad_choices(
             state,
             no_choices,
             base_probabilities,
+            skip_failed_choices,
```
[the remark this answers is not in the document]
i-am-sijia: addressed

activitysim/core/workflow/state.py
```diff
@@ -939,6 +939,62 @@ def add_table(
             # at some later time if desired.
             self.existing_table_status[name] = True
         self.set(name, content)
+        # TODO: do not update tables if no new households were skipped
+        # right now it is updating every time which is inefficient
+        if self.get("num_skipped_households", 0) > 0:
+            self.update_table()
```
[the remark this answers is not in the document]
i-am-sijia: Thanks copilot. As I noted in the TODO, I've added the code to update tables only when new households are being skipped.

activitysim/core/util.py
```diff
@@ -683,6 +683,9 @@ def drop_unused_columns(
 
     unique_variables_in_spec |= set(additional_columns or [])
 
+    # always keep household_id
+    unique_variables_in_spec.add("household_id")
```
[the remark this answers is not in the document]
i-am-sijia: This is fine, no changes needed. As in line 737 below, we are keeping columns in chooser table IF they exist in the `unique_variables_in_spec`.

activitysim/core/workflow/state.py
```diff
@@ -939,6 +939,62 @@ def add_table(
...
+
+    def update_table(self, name: str = None):
```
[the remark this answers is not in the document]
i-am-sijia: This method has been substantially updated to implement a few additional features.

activitysim/abm/models/util/school_escort_tours_trips.py
i-am-sijia: Added the warning for item 1. For item 2, I implemented the check in core.workflow.state.update_table(), immediately after the households are being dropped to make sure it raises an error when N (or the weighted N) exceeds the threshold. Implementing the check in the state instead of in model components makes the code base clean, i.e., less code duplication. The global parameter of the threshold is added in core.configuration.top.fraction_of_failed_choices_allowed.

activitysim/abm/models/location_choice.py
i-am-sijia: Both the warning and the check have been implemented in `core.workflow.state.update_table()`. I implemented them in the state.py instead of in model components to make the code base clean, i.e., less code duplication.

activitysim/abm/models/trip_matrices.py
i-am-sijia: I've moved the summary of number of skipped households, total and total by model component, to `cli.run.run()`, which is a better home than trip_matrices.py.

activitysim/abm/models/trip_matrices.py
i-am-sijia: I've updated the adjustment to expansion weights to be based on the weight of households instead of counts. I agree with you in theory that it should be based on weights. However, I don't think it will make a difference in application. Here's why. In simulation, 99% of the time ActivitySim has the same [sample rate](https://github.com/ActivitySim/activitysim/blob/aa58cc453e1a0ff5c7d5498776f2f594b6085658/activitysim/abm/tables/households.py#L98-L104) for all households and the same expansion weight (calculated as [1/sample rate](https://github.com/ActivitySim/activitysim/blob/aa58cc453e1a0ff5c7d5498776f2f594b6085658/activitysim/abm/models/trip_matrices.py#L415-L418)), this is especially true when ActivitySim uses an input synthetic population with weight of 1 for each record implicitly built in. In contrast to simulation, in estimation, the input households are very likely to have different weights, however, we don't use weights when estimating a model, and `trip_matrices` step is usually skipped in the estimation mode.

activitysim/abm/models/trip_mode_choice.py
i-am-sijia: Yes, you're right. Updated.

activitysim/core/configuration/top.py
i-am-sijia: Added `fraction_of_failed_choices_allowed` as a global parameter and implemented the checking in `core.workflow.state.update_table()`

activitysim/core/logit.py
i-am-sijia: Updated to `warning` for all loggings of skipping households.

i-am-sijia:
@jpn-- Thank you for your review and comments! I've made further updates. Primary changes include:

- Added a `fraction_of_failed_choices_allowed` parameter that users can optionally specify in the `settings.yaml` to set the **threshold** of failed/skipped households (relative to sample households) above which the run should terminate. I updated the test to incorporate this as well.
- Added code to summarize and print out number of skipped households **by model component** at the end of the run.
- Added code to update the state tables only when there're **new** households skipped, improving efficiency.
- Updated code to adjust trip matrix expansion factor using household **weight** instead of count. For this item, please see my response comment to your original comment.
- As requested by @bhargavasana, I added a feature to automatically trace one failed choice per model component. This turned out to be more complex than I expected. I couldn't find an clean way to do it without significant refactoring of the current tracing mechanism and the utility calculation. I managed to implement a workaround for simple simulate models by dynamically adding trace IDs to the tracing object, and re-simulate a single failed choice to trace the utility calculation with minimal runtime impact runtime. For the interaction simulate models, users can manually add the failed household ID to `trace_hh_id` for debug tracing.
- Added documentation in the user guide.

docs/users-guide/ways_to_run.rst
```diff
@@ -283,3 +283,35 @@ With the set of output CSV files, the user can trace ActivitySim calculations in
 help debug data and/or logic errors.
 
 Refer to :ref:`trace` for more details on configuring tracing and the various output files.
+
+.. _skip_failed_choices_ways_to_run :
+
+Skip Failed Choices
+______________________
+
+By default, ActivitySim will skip any failed choices during model execution, i.e., ``skip_failed_choices`` is set to ``True``.
```
dhensle, line 292: I would argue that we want the default behavior set to False.  Having non-expert users just running ActivitySim and dropping households could really confuse folks and cause outputs to not make sense.  Imagine, for example, a planning practitioner is trying to run a parking cost scenario -- they update the parking costs but miss a zone, that zone propagates NAs, and all of a sudden, they have 10% less households in the output file than the input file.  Run says it succeeded and they think all is groovy.  But their revenue forecast is now off by 10% and they don't even know!

activitysim/abm/models/trip_matrices.py
```diff
@@ -90,6 +90,34 @@ def write_trip_matrices(
...
 
+    # This block adjusts household sample rate column to account for skipped households.
+    # Note: the `HH_EXPANSION_WEIGHT_COL` is pointing to the `sample_rate` column in the households table.
+    # Based on the calculation in write_matrices() function, the sample_rate is used to calculate the expansion weight as 1 / sample_rate.
+    # A sample_rate of 0.01 means the sample household should be expanded 1/0.01 = 100 times in the actual population households.
+    # In simulation, the `sample_rate` is calculated and added to the synthetic households
+    # based on household_sample_size / total_household_count, and therefore is the same for all households.
+    # In estimation, the `sample_rate` may vary by household, but weights are not used in estimation, and write_trip_matrices is not called during estimation.
+    # But we still try to cover both cases (when rates are the same vs when they vary) here for consistency.
+    hh_weight_col = model_settings.HH_EXPANSION_WEIGHT_COL
```
dhensle, line 101: Note that the sample_rate is set to only 3 digits of rounding.  This means that for cases where we are running, say, only 100 households to test some spec development on a region with 1.2M households, you will likely get a divide by 0 error here. I suggest we increase the rounding to something much more than 3 and also put in a bit of divide-by-zero protection in this block.

activitysim/core/workflow/state.py
```diff
@@ -938,8 +938,181 @@ def add_table(
...
+                    assert skipped_hh_df.index.get_level_values(
+                        "household_id"
+                    ).is_unique, "household_id is not unique in households_skipped"
+                self.set("households_skipped", skipped_hh_df)
+                # mark households_skipped table as salient and edited
+                self.existing_table_status["households_skipped"] = True
+
+                # Check if we've exceeded the allowed fraction of skipped households
+                # Use weighted households if expansion weight column exists, otherwise use counts
+                if "sample_rate" in df.columns:
```
dhensle, line 1061: See above comment about sample rate.  I think it applies here too.

test/skip_failed_choices/test_skip_failed_choices.py
```diff
@@ -0,0 +1,160 @@
...
+    # set the threshold to 1% of households are allowed to be skipped
+    state.settings.fraction_of_failed_choices_allowed = 0.01
+
+    # the run will fail because more than 1% households will be skipped
+    with pytest.raises(RuntimeError) as excinfo:
+        state.run(models=state.settings.models, resume_after=None)
+    assert "exceeds the allowed threshold of" in str(excinfo.value)
+
+
+def test_skip_failed_choices_high_threshold_prototype_mtc(state):
```
dhensle, line 111: Can we add an additional check here to make sure a trace gets created for the last household that was skipped for the auto ownership and trip mode choice models?

docs/users-guide/ways_to_run.rst
```diff
@@ -283,3 +283,35 @@ With the set of output CSV files, the user can trace ActivitySim calculations in
...
+Users can optionally set a ``fraction_of_failed_choices_allowed`` parameter in the settings file to specify a threshold for the 
+maximum allowable fraction of failed households, this value is expected to be between 0 and 1. 
+If the fraction of failed households exceeds this threshold, ActivitySim will raise a RuntimeError and terminate the model run.
+If the fraction is within the allowable limit, the model will proceed with the skipped households as described above. This threshold
+provides an additional layer of control for users to skip problems when they are small, and stop the model when they are large.
+
+When ``skip_failed_choices`` is enabled, ActivitySim will automatically perform debug tracing for one of the failed households within each 
+model step where failed choices occur. The trace files will be saved in the output/trace directory with folders suffixed by 
+``_resimulate``. This automatic tracing feature allows users to easily investigate the reasons behind the failed choices without needing to 
+manually specify trace IDs. This feature is partially implemented for simple simulate models, and is not yet available for interaction_simulate models.
```
dhensle: What do you mean by "partially implemented for simple simulate models" here?

dhensle (commented):
Looks good! Just a couple of relatively minor suggestions.

docs/users-guide/ways_to_run.rst
bhargavasana: I support this - it might be better have default set to False.

i-am-sijia: As discussed on the engineering call today, this is ultimately a product decision that the product & community team will make. I'm comfortable with either direction. I did want to add a few considerations from similar / different angle:

1. Enabling `skip_failed_choices` could cause existing model run results to change, because it would surface choice failures that are currently masked by the `overflow_protection` feature. I recognize the value of preserving agencies' existing implementations.
2. When `skip_failed_choices` is set to False, the 10% households in David's example still make a choice because `overflow_protection` kicks in and masks the failure. Those households will very likely not be responsive to parking cost changes. The practitioner wouldn't know about this behavior either, and may be confused about why the model appears insensitive. This is a tradeoff worth considering.
3. David is right that the output household is fewer than the input, but that's the intended design, just like how ActivitySim runs on a sample of households. This is reflected in the adjusted `sample_rate` column in the output households, and the resulting trip tables and matrices are scaled up accordingly.

activitysim/abm/models/trip_matrices.py
i-am-sijia: That's a good point. I will add that. Thanks!

jpn--: @i-am-sijia I think the divide-by-zero problem has been reduced by setting the rounding to 7 instead of 3, but the problem still potentially exists here; the sample_rate could still in some cases be zero.  Can we do a np.where or similar to solve it?

Also: why are we rounding this at all?  It seems to me it doesn't hurt us to just store and use the value as a float.

i-am-sijia: I think skip rounding and directly use the value is the better solution. `sample_rate` is a `float64` regardless of the number of decimals.

test/skip_failed_choices/test_skip_failed_choices.py
i-am-sijia: added.

docs/users-guide/ways_to_run.rst
i-am-sijia: It is fully implemented for simple simulate. Corrected wording.

i-am-sijia:
@jpn-- @dhensle , I've implemented divide-by-zero protection for `sample_rate` as discussed at https://github.com/ActivitySim/meeting-notes/issues/79. Please give it a review and let me know if you have any comments.

Software Behavior:
1) **When `sample_rate==0`  for _all_ households**
This can occur when simulating a very small household sample (e.g., ~20 out of 2M) and a computed `sample_rate` is rounded down to zero. 
In this case, the adjustment factor is calculated using household counts instead of weights.

2) **When `sample_rate==0` for _some_ households (partial zeros)**
This can occur when `sample_rate` is provided in the input household data and includes zeros (unclear whether intended or a data error).
In this case, the adjustment factor is computed using only non-zero-sample_rate households (weights from non-zero-sample_rate households only; zero-sample_rate households are skipped). Trips from zero-sample_rate households are skipped when writing out trip matrices.

As discussed, we will remove the rounding of `sample_rate` in a separate PR, with updates to our testing data. I've reverted rounding to 3 decimals as it originally was, to pass test validations.

docs/users-guide/ways_to_run.rst
bhargavasana: #1052 created to get input from @ActivitySim/product.

jpn-- (approved):

dhensle (approved):
Looks good, thanks Sijia!

Issue #595, opened 2022-09-02
Problem in CDAP estimation mode


jpn--:
As reported by @jfdman 

It appears that the coefficients for 3-person households and 4-person households are not being added to the right utility equations in estimation mode. These are the coefficients that are applied when the alternative is where all persons regardless of person type choose the same pattern. They are expressed as P_coeff_D_xxx, P_coeff_D_xxxx, and P_coeff_D_xxxxx, where D is either M,N, or H and the x’s refer to the number of persons in the household.
 
Here is the utility trace for alternative 1 (MMM) for a 3-person household. It looks like the all-household member interaction term for 5-person households ( P.coef_M_xxxxx) is being applied to this alternative instead of the interaction term for 3-person households ( P.coef_M_xxx). 

```
+ P.coef_full_time_worker_asc_M * X('p1_ptype == 1')
+ P.coef_part_time_worker_asc_M * X('p1_ptype == 2')
+ P.coef_university_student_asc_M * X('p1_ptype == 3')
+ P.coef_UNAVAILABLE * X('p1_ptype == 4')
+ P.coef_UNAVAILABLE * X('p1_ptype == 5')
+ P.coef_driving_age_child_who_is_in_school_asc_M * X('p1_ptype == 6')
+ P.coef_pre_driving_age_child_who_is_in_school_asc_M * X('p1_ptype == 7')
+ P.coef_pre_driving_age_child_who_is_in_school_interaction_with_age_6_to_9_M * X('(p1_ptype == 7) & (p1_age >= 6) & (p1_age <= 9)')
+ P.coef_pre_driving_age_child_who_is_in_school_interaction_with_age_13_to_15_M * X('(p1_ptype == 7) & (p1_age >= 13) & (p1_age <= 15)')
+ P.coef_pre_driving_age_child_who_is_too_young_for_school_asc_M * X('p1_ptype == 8')
+ P.coef_pre_driving_age_child_who_is_too_young_for_school_interaction_with_age_0_to_1_M * X('(p1_ptype == 8) & (p1_age >= 0) & (p1_age <= 1)')
+ P.coef_pre_driving_age_child_who_is_too_young_for_school_interaction_with_age_4_to_5_M * X('(p1_ptype == 8) & (p1_age >= 4) & (p1_age <= 5)')
+ P.coef_full_time_worker_interaction_with_age_less_than_40_M * X('(p1_ptype == 1) & (p1_age < 40)')
[...]

PR #1026, opened 2025-12-11
Repair CDAP estimation

jpn--:
This pull request fixes an error in Larch estimation of the CDAP model, reported in #595.  There was a bug in the code, where interaction terms for 3, 4, and 5 person households adopting an all-M, all-N, or all-H patterns were all mapped to the same coefficient; the original model structure has unique parameters for each household size. [[1]](diffhunk://#diff-80c8e46b25760ec13baffd65aa7fce997341d2c8e2e2ba42a1010bf3d9ad66feL177-R206) [[2]](diffhunk://#diff-80c8e46b25760ec13baffd65aa7fce997341d2c8e2e2ba42a1010bf3d9ad66feR225-R239)

### Testing and validation

* Expanded the `test_cdap_model` test to include assertions verifying the correct assignment of interaction parameters, specifically checking the presence and absence of wildcard coefficients for different model sizes.
* Updated the regression test data in `test_cdap_model_loglike.csv` to reflect changes in log-likelihood values resulting from the repaired coefficient mapping logic.


docs/dev-guide/changes.md
```diff
@@ -8,39 +8,56 @@ configurations or code to fail to run correctly.
...
+entire household adopting the same daily activity pattern (M, N or H).  However,
+due to a coding error, the same parameter was being used across all three household
+sizes.  This has now been fixed, and the correct interaction terms are now specified
+for each household size.  Users who re-estimated CDAP parameters using Larch with
+prior versions of ActivitySim may want to review and/or re-estimate any CDAP
+parameters to ensure that they are correct.  It is not expected that this error
+will have a significant impact on simulated travel behavior outcomes, as it is only
+a few parameters with a large and complex model, and the overall behavioral
+characteristics of 3, 4, and 5-person households are expected to be somewhat
+similar to each other in any case.
```
dhensle: I believe the fix would be to actually be adding a `coef_[M,N,H]_xxx `and `coef_[M,N,H]_xxxx` and giving them the same value as `coef_[M,N,H]_xxxxx` since these were all combined in estimation, right?
Granted, if the model had undergone some sort of calibration, that difference would have been calibrated out and they should probably leave it.

So, I think we should supply the following guidance:
1. Make the new coefficients if you have estimated CDAP but not calibrated.
2. Just leave it if calibrated -- minor bug with very limited sensitivity differences.

activitysim/estimation/test/test_larch_estimation/test_cdap_model.csv
```diff
@@ -1,162 +1,168 @@
...
-coef_N_77,2.3038032526813228,2.3038032526813228,1.5529999999999999,0
-coef_N_78,-0.050192112434025701,-0.050192112434025701,0.61839999999999995,0
-coef_N_88,-0.12160277921446122,-0.12160277921446122,0.87709999999999999,0
-coef_N_xxxxx,-1.0454032140665002,-1.0454032140665002,-3.4529999999999998,0
+coef_N_77,2.3558560142647202,2.3558560142647202,1.5529999999999999,0
+coef_N_78,-0.29812068125334135,-0.29812068125334135,0.61839999999999995,0
+coef_N_88,-0.05952606764773704,-0.05952606764773704,0.87709999999999999,0
+coef_N_xxx,-0.89914972777069246,-0.89914972777069246,-0.36530000000000001,0
+coef_N_xxxx,-0.72130413279127681,-0.72130413279127681,-1.3460000000000001,0
+coef_N_xxxxx,-6.6290578146794195,-6.6290578146794195,-3.4529999999999998,0
```
dhensle, line 113: Happy to see the additional coef_*_xxx and coef_*_xxxx coefficients.  All these coefficients across M,N, and H make a lot more sense now with a strong negative coefficient for everyone in a 5-person household doing the same thing.

activitysim/estimation/larch/cdap.py
```diff
@@ -174,8 +199,11 @@ def cdap_interaction_utility(model, n_persons, alts, interaction_coef, coefficie
             c_split = c.split("_")
             for j in c_split[2:]:
                 interact_coef_map[(c_split[1], j)] = c
-                if all((i == "x" for i in j)):  # wildcards also map to empty
-                    interact_coef_map[(c_split[1], "")] = c
+                # previously, wildcards also mapped empty here, but this caused a clash
+                # as all wildcards would map to the same coefficient name not matter the
```
dhensle: minor typo here, should be "name no matter" instead of "name not matter"

dhensle (approved):
Fix looks good.  Suggest we add more detail to the change log, and one minor typo to fix.

docs/dev-guide/changes.md
jpn--: I've added a note here to this effect.

PR #1027, opened 2025-12-18
use all TAZs when using sharrow

### LINKED ISSUE #1022: 'utility not aligned' error when running sharrow in test mode using uv, but not in conda using same version of Activitysim.
**Describe the bug**
This is an odd one. When running sharrow in test mode with a uv environment, I get a 'utility not aligned' error [here](https://github.com/ActivitySim/activitysim/blob/main/activitysim/core/simulate.py#L812), but not when running the same version of activitysim using anaconda. I did some debugging and the utilities that are causing the issue are coming from the bike and walk tour mode choice utilities when getting logsums during mandatory_tour_scheduling step, so perhaps this has something to do with the maz_to_maz skims. When using the anaconda env, the utilities from variables utilities and sh_util are aligned. Interestingly, when switching to an older version of Activitysim the bug turned up sooner in the disaggregate accessibilities step.  I tested this against an older anaconda environment and a new one. For the new one,  I first created the environment and then ran "pip install -e" from the (cloned) activitysim directory. This essentially installed activitysim in editable mode and then all of its dependencies using pip.

**To Reproduce**
1. Copy conifgs and input data from [here](https://drive.google.com/drive/folders/1bMUc0Ue4Tmn4Xy5BCARGKuHNDsOH-HEY?usp=sharing).
2. Clone Activitysim
3. Open a command prompt, cd to Activitysim, then enter _uv sync --locked_ to create the uv .venv.
4. Enter _.venv\Scripts\activate_ to activate the uv environment. 
5. Enter _activitysim run -c path_to_configs_sh_compile -c path_to_configs_dev -d path_to_data_test -o path_to_output_
6. The model should crash during mandatory tour scheduling. 
7. Create a new anaconda env by using the environment.yml file in the google drive linked above or by creating a new env and then running _pip install -e_ in the activitysim root dir. 
8. Run the model using the anaconda env using the same arguments as above. 
9. The model should not crash. 

**Additional context**
This is probably a library version issue and not anaconda/uv issue. Either way it will be good to know why the utilities are different running one way and not the other.

jpn--:
This pull request updates the trip destination model to improve compatibility with the "sharrow" mode by ensuring that all TAZs (Traffic Analysis Zones) are represented in aggregated data structures, even if some TAZs have no associated MAZs (Micro Analysis Zones). This change helps align the model's alternatives with TAZ-based skim matrices, which is particularly important when using the sharrow framework.

**Sharrow compatibility and TAZ completeness:**

* Updated `aggregate_size_term_matrix` to accept an optional `all_tazs` parameter, ensuring that the resulting TAZ size term matrix includes all TAZs, filling in zeros for any TAZs without MAZs.
* Modified `destination_presample` to pass the complete set of TAZs to `aggregate_size_term_matrix` when sharrow is enabled, guaranteeing all TAZs are present in the matrix.
* Ensured that the `alternatives` DataFrame in `destination_presample` is reindexed to include all TAZs with zero fill when sharrow is active, so that the alternatives vector aligns with the TAZ skim structure.

Closes #1022.


jpn--:
In writing a unit tests for this PR, I discovered a second problem: for two zone systems, the`dtype` for imported MAZ-to-MAZ data was not being controlled to match the `dtype` of the similarly named TAZ-to-TAZ skim, e.g. the MAZ data could be stored as float64 while the TAZ data is stored as float32.  This has minor impacts on runtime, but can cause subtle bugs in handling 2-zone skim data with sharrow with the "fastmath" option turned on.  This is not the first time that "fastmath" has tripped us up (there is already a [note](https://github.com/ActivitySim/activitysim/blob/61078c7198b46c4d4ff336caa7678bd814b6fa82/activitysim/core/configuration/base.py#L171-L172) in the documentation about turning it off when finding these exact kinds of errors.  It is super hard to diagnose these problems when they do occur so I decided the best course is to change the default on "fastmath" to False.  Users can easily turn it back on for individual components when the performance gains are useful.

Issue #820, opened 2024-02-26
Some zones sometimes reopen while using shadow pricing simulation


jpn--:
**Describe the bug**
In investigating #815, I discovered that the shadow pricing simulation algorithm sometimes re-opens zones that were previously closed.

<img width="547" alt="image" src="https://github.com/ActivitySim/activitysim/assets/1036626/87f4b62b-15e7-407b-a2e4-57bd8c9d74cb">

**To Reproduce**

- Use main branch @ff3e96b.
- Run full size prototype mtc extended model with 100% household.
- Trace hh id 1098914
- Compare `shadow_price_utility_adjustment` column from `trace/school_location/i*/sample/interaction_sample/alternatives-*.csv`

**Expected behavior**
Once the shadow price simulation algorithm closes a zone, it should stay closed.  Re-opening a closed zone probably doesn't completely void the results here -- in all cases I observe, it then gets closed again on the next iteration, but it probably prolongs convergence as workers or students will certainly just re-oversubscribe that zone again.

dhensle:
The simulation-based constraint algorithm selects people to re-simulate based on the probability of `sprice = 1 - (desired_share / modeled_share)`, but the "shadow price" of the zone set to -999 is calculated as `np.where((sprice <= 1 + percent_tolerance / 100), -999, 0)`. (See it in the code [here](https://github.com/ActivitySim/activitysim/blob/main/activitysim/abm/tables/shadow_pricing.py#L868).)  

Notice how the percent tolerance shows up in turning the zone on-and-off, but not in the selection of people to re-simulate.  I _think_ this is the source of the discrepancy....  In other words, I can have a 5% tolerance, with a zone that has 100 enrollment.  If 103 students select that zone, the sprice is > 1 and I can choose people from the zone to re-simulate, but it is within the percent tolerance and the shadow price is -999.  If the 3% probability selection just so happens to choose more than 3 people to resimulate, then the zone will open back up in the next iteration.

If my logic is correct here, I see a couple ways forward: 
1. Just leave it as is. I don't think this is necessarily a "bug" but rather a side effect of having a tolerance.
2. Update the sampling logic to say that if people are within the tolerance for the zone, they will just remain there.

cc: @aletzdy, @jfdman

bettinardi:
I like option 2

joecastiglione:
I agree that option 2 seems cleaner, and consistent with the purpose of
having the tolerance setting in the first place.

Joe

[...]

PR #1028, opened 2025-12-30
Fix for zones reopening in simulation-based constraint mechanism for work/school location choice

dhensle:
Fix for #820:  If zones close, they now remain closed.


jpn-- (approved):
This is great!  Here's why I like it:

- You have written a clear and simple test that creates a tiny model, runs it, and fails on the existing code in a manner that clearly demonstrates the existing problem and runs in under one second.
- After the code changes are made, the test passes.  Voila!

The one thing that's missing is a note on the change log `activitysim/docs/dev-guide/changes.md`.  This is not really what may of us would call a "breaking change" as it should not formally cause any existing model to fail, but it does break strict reproducibility between versions.  This is seen in the SANDAG test, which now fails because the shadow-priced location choices are changing just a tiny bit.

dhensle:
Great! Glad you like the test -- it was not a trivial exercise to get a nice CI test built for this particular issue. And we also now have a CI test for the simulation-based constraint mechanism as a whole as a bonus!

I have updated the change log.  I think this is ready to merge!

jpn-- (approved):

PR #1029, opened 2026-01-22
explicit error terms updated post Activitysim 1.5.1

m-richards:
VLC is doing some work with DTP Victoria to test their activitysim model with EET functionality. They've recently rebased to 1.5.1, and so need EET working with the 1.5.1 code for this test. As I understand the consortium is still interested in pursuing developments with EET, I thought we could share this rebasing in case it's a useful starting point.

I had originally intended to keep this on our fork for the moment with the intent of opening a pull request here, but I've inadvertedly created one on the upstream already. Given this is built on work already done by jan across several commits and years, I've opted to merge up to date with main, rather than rebase my way through all the historical conflicts.

Note that the diff here is very large - you'll see a much cleaner diff with the base set to ActivitySim:main instead of ActivitySim:explicit_error_terms, but the intention here was to try and update that branch, rather than supercede it.

Edit: It's probably also worth noting that as part of this work, we've identified a few additional spots where EET doesn't seem to be fully implemented (where logit.make_choices is called directly, without wrapping) but we haven't addressed this in this merge request:
* CDAP extra hh member choices (households with > 5 people)
* Non mandatory tour frequency second step - extend tour counts
* vehicle type choice second step - fuel type
* trip purpose
* trip scheduling
* shadow pricing simulation method
* tour scheduling probabilistic


janzill:
This looks good, thanks Matt! @jpn-- happy for this to be merged into the EET branch.

m-richards:
superceded by #1042, have discussed the notes in here about extra places to apply EET with Jan.

Issue #1038, opened 2026-02-18
ActivitySim Testing Plan


jpn--:
## Background and Goals

ActivitySim's current testing regime relies heavily on end-to-end integration tests that run entire model pipelines and check for bitwise-identical outputs. While these tests provide meaningful coverage, they have two serious drawbacks: they're slow (CI runs take an hour or more), and they're brittle — any small code change tends to break them, triggering investigation and manual correction of expected outputs rather than actual debugging. The goal of this plan is to shift the balance toward faster, more targeted unit tests, while preserving the integration tests that catch real-world regressions. Critically, this shift needs to happen incrementally, without a large upfront investment that would be difficult to justify to funders.

## The Problem with ActivitySim's Current Testing Approach

ActivitySim's testing infrastructure has grown organically alongside the codebase, and it shows. The result is a system that is simultaneously slow, brittle, and incomplete — a combination that creates real friction for developers and quietly erodes confidence in the codebase over time.

The dominant form of testing today is what might be called "implementation testing": running an entire model pipeline from start to finish and checking that the outputs are identical to a previously saved reference. These tests are valuable in one sense — they do catch regressions — but they come with serious costs. A full CI run takes an hour or more, which means developers often wait a long time only to discover a problem they could have caught much earlier. Worse, because the tests check for bitwise-identical outputs, even a trivially small change to the code — adjusting a rounding behavior, reordering an operation, fixing an unrelated bug — can cause a cascade of test failures. These failures don't indicate that something is wrong with the model; they indicate that something changed. Distinguishing between the two requires manual investigation, and correcting the reference outputs is tedious work that adds no real value. The practical effect is that developers learn to dread test failures rather than trust them.

The tests are also poorly organized. They are spread across multiple locations without a clear rationale, and there is no obvious place to look when trying to understand what is and isn't covered. This makes it hard for a new contributor to know where to add a test, or even whether a test already exists for the thing they're changing. The path of least resistance is to not write a test at all.

Underlying all of this is a structural gap: there are few unit tests. Many individual functions and components are not tested in isolation, which means bugs can hide in plain sight as long as the end-to-end outputs happen to remain stable.
The end-to-end nature of testing also makes it difficult to test various different input settings. If one particular feature of a model has three different possible settings, and another feature that it interacts with can be turned on or off, running end-to-end tests require 3 * 2 = 6 different full model runs to exercise those interactions.
[...]

PR #1041, opened 2026-03-10
ENH: add trip scheduling choice explicit chunking

### LINKED ISSUE #1040: Trip scheduling choice doesn't support explicit chunking
**Describe the bug**
I wanted to check and see if there's a reason that trip scheduling choice was never wired up with the explicit chunking mechanism when it was added (happy for this not to be called a bug if this was deliberate!). In the DTP model I've been looking at, this step is now the memory peak (after explicit chunking has been configured in the other steps).

Happy to split out a PR since I think this is trivial to wire up, but thought I'd see if there's a good reason not to first. I can see that explicit chunk size is also not fed through in trip purpose, `initialize_tvpb`, `run_tour_scheduling_probabilistic` and `simple_simulate`, maybe there was just a pragmatic choice about what was likely enough to be memory bottlenecks? Though these all would have been handled under the old training/ production mode options I think?

```python
# run_trip_scheduling_choice snippet
if len(indirect_tours) > 0:
        # Iterate through the chunks
        result_list = []
        for (
            i,
            choosers,
            chunk_trace_label,
            chunk_sizer,
        ) in chunk.adaptive_chunked_choosers(state, indirect_tours, trace_label):
            # Sort the choosers and get the schedule alternatives
            choosers = choosers.sort_index()
            schedules = generate_schedule_alternatives(choosers).sort_index()
```
the adaptive chunking helper doesn't feed through a chunksize, and the respective pydantic class `TripSchedulingChoiceSettings` also doesn't have a key to set the chunk size paramter.

m-richards:
Closes #1040

Based on discussion in #1040 - this model is still being used in the DTP activitysim implementation and with certain other tweaks on their activitysim fork, this step will out of memory on 128GB hardware.

This is pretty simple to wire up. I've added a basic test which shows the parameter is accepted (but it would also pass if I hadn't wired up chunk.adaptive_chunked_choosers, so it's not ideal)

I'm not quite sure what the testing philosophy is at the moment - I've read #1038 and the google doc, not exactly sure what's pragmatic with a small change like this - given there aren't a host of other tests testing for explicit chunking.


activitysim/abm/test/test_misc/test_trip_scheduling_choice.py
```diff
@@ -285,3 +286,24 @@ def test_run_trip_scheduling_choice(model_spec, tours, skims, locals_dict):
 
     # check that tours with no inbound stops have zero inbound duration
     assert out_tours[tsc.IB_DURATION].mask(in_tours[tsc.HAS_IB_STOPS], 0).sum() == 0
+
+    # confirm explicit chunking is supported and doesn't affect results
```
m-richards, line 290: I can convert this to a parametrised test in the same way as test_agg_accessibility_orig_land_use, I'm not sure if that's preferred or not

jpn--:
Closed by #1102

PR #1062, opened 2026-03-25
Implement deletion of temporary variables in summarize.py

ziocolaandrea:
While running the summarise step I ran into memory allocation issues.
I tested adding the deletion of temporary variables not used in subsequent summaries and managed to go through the step.


dhensle:
These review comments as well as documentation are addressed in #1116

Issue #1052, opened 2026-03-11
Default setting for skipping households on simulation failure


jpn--:
We are about to implement in #1023 a great new feature: the ability to allow a limited number of failed choices in a model component, which will warn and log errors, and remove failed households from further processing, but not stop a model run.  There is some debate over whether the default setting for this feature should be True or False.  How we implement this has been solved by @ActivitySim/engineering but whether the default setting should be True or False needs guidance from @ActivitySim/product 

if `True`:

- This great new feature will be active by default.
- Users who previously built models that rely on overflow protection will have results from those models now change.  The most likely scenario here is that overflow protection is masking places where the utility function is trying to make all alternatives unavailable by applying large negative values.  Currently the system is forcing a choice on those cases, typically an alternative with the smallest large negative penalty.  This will now manifest as a failed choice.  
- Some models will drop households. This could really confuse folks and cause outputs to not make sense. Imagine, for example, a planning practitioner is trying to run a parking cost scenario -- they update the parking costs but miss a zone, that zone propagates NAs, and all of a sudden, they have 10% less households in the output file than the input file. Run says it succeeded and they think all is groovy. But their revenue forecast is now off by 10% and they don't even know!

if `False`:

- ActivitySim will give the same results as previously unless the user takes the initiative to turn on the new feature.
- Since users will need to take action to enable this feature, they perhaps will be more informed about the consequences.

DavidOry:
@jpn-- 
Is there currently a switch to turn the overflow protection on/off? If so, what is the default for that? 

This approach to allowing model runs to continue with errors present strikes me as better than the overflow protection. So in the long run it would be good to remove the overflow protection. Is that on the roadmap? (Either literal or in our heads?)

Do we have any sense as to the number of users who are currently relying on overflow protection and the scale of the problems this is covering over?

In your hypothetical, you say turning this feature on could result in 10 percent fewer households. If that's true, doesn't that mean there's some error now being covered up by overflow protection that is resulting in bad forecasts? It's not obvious to me that the bad forecasts being covered up by the overflow protection are worse that the forecast being "off by 10 percent", as in your hypothetical. Is it just worse because the current error is the status quo?

Your hypothetical also implies that ActivitySim is not taking any action to re-scale the households up, either via replication of successful records or by returning an updated sample size parameter to the user. Why not? 

My preference would be:

[...]

PR #1063, opened 2026-04-02
Lower default threshold for skipped failed choices to 0.1%

i-am-sijia:
This pull request tightens the default allowed threshold for failed choices and clarifies documentation around the transition to `skip_failed_choices`. 

- It lowers the default `fraction_of_failed_choices_allowed` from 0.1 to 0.001, as discussed in https://github.com/ActivitySim/activitysim/issues/1052#issuecomment-4077461595, causing the model to fail with far fewer skipped households, and 
- Updates the user guide to explain the new behavior, including expected warnings and potential crashes when agencies run their existing model with this new feature for the first time.

**Impact on existing model runs:**
With `skip_failed_choices` defaulted to `True` and an allowed failure threshold as low as 0.1%, failures that were previously silently masked will now generate warnings, be explicitly skipped, and counted toward the threshold. As a result, model runs that previously completed "successfully" may now fail earlier, surfacing underlying specification or data issues that require attention rather than being silently absorbed. @jpn-- we may want to include this note in the next ActivitySim release note.

### DIFF
--- a/activitysim/core/configuration/top.py
+++ b/activitysim/core/configuration/top.py
@@ -790,7 +790,7 @@ class Settings(PydanticBase, extra="allow", validate_assignment=True):
     .. versionadded:: 1.6
     """
 
-    fraction_of_failed_choices_allowed: float = 0.1
+    fraction_of_failed_choices_allowed: float = 0.001
     """
     Threshold for the fraction of households that can be skipped before failing the model run,
     used in conjunction with `skip_failed_choices`.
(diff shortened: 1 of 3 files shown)


jpn-- (approved):
LGTM.

The place to put the notice for users so that it appears in the next release notes is [here](https://github.com/wsp-sag/activitysim/blob/default-max-failed-hh/docs/dev-guide/changes.md). Can you do that and then I will merge.

i-am-sijia:
@jpn-- Added to change log. Thanks!

PR #1042, opened 2026-03-10
update EET feature branch

jpn--:
The PR updates the feature branch to stay mostly in-sync with development on the main branch.

PR #1005, opened 2025-10-21
Trip Scheduling Choice -- Same Results Single Process and Multi-Process

dhensle:
Fix for #891 

This pull request introduces several improvements and refactors to the trip scheduling and departure choice components, settings, and configuration files in the ActivitySim ABM prototype and test suite. The main changes include updating configuration schemas for consistency, refactoring settings classes for better inheritance and maintainability, improving sorting and spec reading logic, and enhancing test coverage and utility functions.

**Configuration and Settings Refactoring**

* Updated the trip scheduling and departure choice configuration files to use `SPEC` instead of `SPECIFICATION`, standardized preprocessor settings, and added `compute_settings` and `protect_columns` for consistency and clarity. (`activitysim/examples/prototype_arc/configs/trip_scheduling_choice.yaml`, `activitysim/abm/test/test_misc/configs_test_misc/trip_scheduling_choice.yaml`, `activitysim/abm/test/test_misc/configs_test_misc/trip_departure_choice.yaml`, [[1]](diffhunk://#diff-a17bba3fe1ee2a1ed3cd4425adf4ae826f621462fd70c613a253d05fc81dda9aL12-R21) [[2]](diffhunk://#diff-851e0196532cba201c2d4f1f04d7b785345f77a7a04ec6d75a998d04d1206c52R1-R28) [[3]](diffhunk://#diff-b5ebe34b7b9267acaa0dc7e1ca3fff84d47aff88240d6d68fb6f30fdc75a2d01R1-R20)
* Refactored the `TripSchedulingChoiceSettings` class to inherit from `LogitComponentSettings` and removed redundant fields, streamlining the settings schema for trip scheduling choice. (`activitysim/abm/models/trip_scheduling_choice.py`, [activitysim/abm/models/trip_scheduling_choice.pyL347-R352](diffhunk://#diff-85380e034bad59819e869a62d7f1e80b1057ba8521d8e82d59c0412df0763bdeL347-R352))

**Logic and Code Improvements**

* **Added sorting by `tour_id` and `SCHEDULE_ID` in the schedule generation function to ensure deterministic and organized output. (`activitysim/abm/models/trip_scheduling_choice.py`, [activitysim/abm/models/trip_scheduling_choice.pyR85](diffhunk://#diff-85380e034bad59819e869a62d7f1e80b1057ba8521d8e82d59c0412df0763bdeR85))** (This was the actual fix to #891)
[...]

PR #1064, opened 2026-04-03
Explicit error terms production code

janzill:
This PR brings the explicit error term (EET) work to production standard. It contains the following major changes compared to the PoC implementation:
	1. It adds testing and documentation for EET
	2. It decouples the sampling method from the simulation method and adds Poisson sampling (based on @m-richards  implementation in #1065), including tests and documentation. It also addresses runtime and memory issues with the sampling method "eet".
	3. It removes several inconsistencies in the EET simulation branch where edge cases could have led to unexpected changes in choices for individual choosers due to non-alignment of error terms.
	4. It consistently implements EET for nested logit models

Note that the default simulation method remains Monte Carlo, with all existing unit and integration tests unchanged and passing. Users should therefore not see any differences in their model runs unless explicitly opting in to EET. This is done by adding `use_explicit_error_terms: True` to the settings. This is a drop-in replacement, with default settings leading to a 3-10% runtime increase of a single demand model run for the models we have tested so far.

This is a large PR and I will add details in comments below so we can keep discussions focused.


janzill:
Regarding tests:

- Regarding integration tests with the two external models, we cannot compare outcomes between MC and EET due to the small household sample sizes.
- We included EET functionality in semcog tests and noticed a small difference for runs with 1 and 2 processes. Debugging revealed an edge case for probabilistic scheduling. This is independent of EET, the model just looks up probabilities from provided tables and therefore does not have an EET equivalent. It turns out that how failed trips are grouped together can lead to very small differences between single and multi-processing runs for small test sets. This applies to MC as well. Changing the base seed did not trigger that edge case, but worth discussing during the engineering meeting.
- We noticed that ARC tests are disabled but wanted to add these back in because it's the only test model that uses trip_scheduling_choice afaik. It turns out the regress trips had small differences for trip schedules when running with MC. Git history shows the test was disabled in #840. It looks like the test files were later touched in #1005 but the test was never re-enabled. I cannot see why the test was disabled in the first place, I might be missing some context here but I decided to update the MC regress file and re-enable the test, as well as add one for EET

janzill:
Inconsistencies fixes:

- Improved consistency of error terms during sampling by aligning random draws to the universe of alternatives, not just zero-attraction alternatives
- Improved consistency of error terms for choices from sampled sets by aligning random draws to the universe of alternatives, not the position in the sampled set
- Improved consistency of two-zone pre-sampling by aligning MAZ choices to the universe of alternatives.
- For shadow pricing, choosers now have consistent error terms for sample and final choice over shadow pricing iterations. Note for shadow pricing method simulation, fixing random numbers per chooser over iteration led to a markedly different solution (much longer distances) than without. This was independent of the simulation method and also of the sampling method and purely due to having the same random numbers in the shadow pricing decisions. We therefore introduced a separate RNG channel to keep results in line with current release code. This RNG is only used if use_explicit_error_terms = True and shadow pricing method = simulation.

janzill:
Regarding Poisson sampling and disaggregate accessibilities:

For the MTC_extended and SANDAG models, we saw large differences (about 50%) when running with Poisson sampling compared to the other two methods. The other two methods agree. Disaggregate accessibilities are destination choice logsums, generated by sampling a subset of alternatives, and we traced the differences back to how the sampling correction factor is specified. The bottom line is that in the current specification Poisson is un-biased, whereas both MC and EET sampling are biased by log(sample_size). This is material (on the order of half the mean value) and leads to significant differences in downstream models that use disaggregate accessibilities, like auto_ownership and cdap in SANDAG. As mentioned in this week's meeting, we will present more details on this in the engineering meeting on 5/21.

janzill:
Regarding nested logit:

The PoC implementation had a recursive structure, with choices made by walking down nests and choosing at each level. The use of logsums means that there were edge cases where unintuitive changes could happen (nest switching), and the recursive structure was slow. We replaced this with a method that draws nested logit error terms directly, solving both of these problems.

activitysim/abm/models/joint_tour_participation.py
```diff
@@ -131,6 +131,7 @@ def participants_chooser(
     choosers: pd.DataFrame,
     spec: pd.DataFrame,
     trace_label: str,
+    nest_spec: Optional[dict, LogitNestSpec] = None,
```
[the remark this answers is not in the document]
janzill: fixed

activitysim/core/random.py
```diff
@@ -445,7 +654,38 @@ def get_channel_for_df(self, df):
...
+            pipeline step name for this step
+        df : pandas.DataFrame
+            df with index name and values corresponding to a registered channel
+        """
+        channel = self.get_channel_for_df(df)
+        channel.row_states.loc[df.index, "offset"] = 0
+        logger.info(
+            f"RNG: resetting random number generator offsets for channel '{channel.channel_name}' for {len(df)} rows"
+            + f" with index name '{df.index.name}'. Total lenght df: {len(channel.row_states)}"
+        )
```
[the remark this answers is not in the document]
janzill: fixed

docs/users-guide/sampling-methods.rst
```diff
@@ -0,0 +1,52 @@
...
+Practical differences:
+
+* ``monte_carlo`` and ``eet`` both sample with replacement, so duplicated sampled alternatives
+  are possible and their aggregate sampled shares track repeated-draw MNL behavior more closely.
+* ``poisson`` samples alternatives by inclusion probability, so each sampled alternative appears
+  at most once per chooser. This can change raw sampled shares in highly peaked cases, even though
+  the downstream sampling correction remains well defined.
+* ``monte-carlo`` is the fastest method, followed by ``poisson``, with ``eet`` being the slowest.
+  However, for models like location choice, most runtime comes from logsum calculations and the
+  total difference between ``monte-carlo`` and ``poisson`` sampling is usually very small.
```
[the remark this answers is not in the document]
janzill: fixed

docs/dev-guide/sampling-methods.md
```diff
@@ -0,0 +1,179 @@
...
+run because it requires many more random draws and more complex logic to avoid materializing large
+chooser-alternative arrays of error terms in memory.
+
+### Poisson Sampling
+
+`poisson` does not perform repeated draws with replacement. Instead, each chooser-alternative
+pair is sampled independently with inclusion probability $1 - (1 - p)^s$, where $p$ is the original
+choice probability and $s$ is the configured sample size.
+A single inclusion draw is made for each alternative. This is much cheaper than repeated draws for
+`eet`, and unlike ``monte-carlo``, it can still benefit from stable alignment of random draws to
```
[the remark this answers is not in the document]
janzill: fixed

docs/dev-guide/sampling-methods.md
```diff
@@ -0,0 +1,179 @@
...
+1 - (1 - p)^s
+$$
+
+Since `pick_count` is always `1` for `poisson`, the correction becomes $\log(1 / \text{prob})$.
+
+This means that all three methods use the same correction expression,
+`np.log(df.pick_count/df.prob)`, even though `prob` has a different interpretation for `poisson`
+than for the with-replacement methods.
+
+<<!-- TODO: Add section on disaggregate accessibilities here once decision is made in engineering meeting -->>
```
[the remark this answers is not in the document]
janzill: fixed

docs/dev-guide/explicit-error-terms.md
```diff
@@ -0,0 +1,181 @@
...
+maximum. For models where this is not the case, ActivitySim can automatically perform the
+conversion for internal calculations. The `recode_columns` option creates contiguous zero-based IDs
+where needed; see the
+[Zero-based Recoding of Zones](using-sharrow.md#zero-based-recoding-of-zones) section for details.
+
+For models that use sub-sampling of alternatives, it is important to keep the sampling scheme
+identical between scenarios, otherwise the error terms for the choice from the sampled set are
+not guaranteed to be aligned. When running with EET, the default sampling method is ``poisson``,
+which balances runtime performance and noise reduction. For more details on sampling methods,
+see {ref}`sampling_methods_dev_guide`.
```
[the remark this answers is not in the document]
janzill: fixed

activitysim/core/interaction_sample_simulate.py
```diff
@@ -496,6 +520,12 @@ def interaction_sample_simulate(
     explicit_chunk_size : float, optional
         If > 0, specifies the chunk size to use when chunking the interaction
         simulation. If < 1, specifies the fraction of the total number of choosers.
+    alts_context: AltsContext, optional
+        Representation of the full alternatives domain (min and max alternative id)
+        in the absence of sampling.
+        This is used with EET simulation to ensure consistent random numbers across the whole alternative set
+        ( as the sampled set may change between base and project). When not provided,
+        EET with integer-coded choice ids will raise an error.
```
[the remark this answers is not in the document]
janzill: fixed

janzill:
@dhensle target is now main as discussed earlier. I cannot resolve the copilot suggestions, but they are all fixed.

activitysim/abm/models/joint_tour_participation.py
```diff
@@ -215,9 +217,31 @@ def participants_chooser(
...
                 # anybody with probability > 0 is forced to join the joint tour
-                probs[choice_col] = np.where(probs[choice_col] > 0, 1, 0)
-                non_choice_col = [col for col in probs.columns if col != choice_col][0]
-                probs[non_choice_col] = 1 - probs[choice_col]
+                if state.settings.use_explicit_error_terms:
+                    # need "is valid choice" such that we certainly choose those with non-zero values,
+                    # and do not choose others. Let's use 3.0 as large value here.
+                    probs_or_utils[choice_col] = np.where(
+                        probs_or_utils[choice_col] > logit.UTIL_MIN,
+                        3.0,
```
dhensle: Suggest we parameterize this 3.0 value in logit.py (named like UTIL_LARGE_ENOUGH or something).  Shows up in interaction_sample_simulate too in the same context.

activitysim/abm/models/trip_scheduling_choice.py
```diff
@@ -279,6 +279,15 @@ def run_trip_scheduling_choice(
         ) in chunk.adaptive_chunked_choosers(state, indirect_tours, trace_label):
             # Sort the choosers and get the schedule alternatives
             choosers = choosers.sort_index()
+            # FIXME-EET: For explicit error term choices, we need a stable alternative ID. Currently, we use
```
dhensle, line 282: Is this fix something that should get just get implemented right now?  Seem better than just leaving this fixme in here for a who-knows amount of time. (The proposed fix make sense to me)

activitysim/core/configuration/base.py
```diff
@@ -135,6 +135,15 @@ class ComputeSettings(PydanticBase):
     Sharrow settings for a component.
     """
 
+    sample_method: None | Literal["monte_carlo", "eet", "poisson"] = None
```
dhensle, line 138: Do we need to be careful about the sampling method for estimation? The actual choice method (eet vs monte-carlo) shouldn't matter because the choice just gets overwritten by the survey choice anyways.  But is the sampling methodology important?

I don't think it should matter, but would like to double check.

activitysim/core/interaction_sample.py
```diff
@@ -17,15 +18,345 @@
...
+    Returns
+    -------
+    sampling_method : InteractionSampleMethod
+    """
+    sample_compute_settings = getattr(model_settings, "compute_settings", None)
+    if sample_compute_settings is not None:
+        sample_compute_settings = sample_compute_settings.subcomponent_settings(
+            "sample"
+        )
+    return _resolve_sample_method(state, sample_compute_settings)
```
dhensle: Combining this into a single resolve_sample_method seems cleaner. I don't think we need another function here.

activitysim/core/interaction_sample.py
```diff
@@ -443,8 +780,45 @@ def _interaction_sample(
 
     state.tracing.dump_df(DUMP, utilities, trace_label, "utilities")
 
-    # convert to probabilities (utilities exponentiated and normalized to probs)
-    # probs is same shape as utilities, one row per chooser and one column for alternative
+    sampling_method = _resolve_sample_method(state, compute_settings)
+
+    # Estimation requires MC sampling and MC choice for now
+    if estimation.manager.enabled and sampling_method != "monte_carlo":
```
dhensle, line 767: If this is true, then we should enforce it in the settings.  But I am not sure why monte_carlo should be required for estimation.

activitysim/core/interaction_sample_simulate.py
```diff
@@ -287,115 +311,165 @@ def _interaction_sample_simulate(
...
+        positions = positions.clip(lower=0)
 
-        # create a series with index from choosers and the index of the chosen alternative
-        choices = pd.Series(choices, index=choosers.index)
+    # resulting pandas Int64Index has one element per chooser row and is in same order as choosers
+    choices = alternatives[choice_column].take(positions + first_row_offsets)
 
-        chunk_sizer.log_df(trace_label, "choices", choices)
+    # resulting pandas Int64Index has one element per chooser row and is in same order as choosers
+    choices = alternatives[choice_column].take(positions + first_row_offsets)
```
dhensle: Remove duplicate call.

activitysim/core/interaction_sample.py
```diff
@@ -17,15 +18,345 @@
...
+    inclusion_probs_values = 1.0 - np.power(
+        1.0 - probs.to_numpy(copy=False), sample_size
+    )
+
+    sampled_values = np.full(inclusion_probs_values.shape, np.nan)
+
+    n = 0
+    active_row_positions = np.arange(len(probs), dtype=np.int64)
+
+    while active_row_positions.size > 0:
```
dhensle: I am not sure I really like how these retries are being applied. I see two issues:

First, if an alternative gets selected in a re-try, I think the inclusion probability correction np.log(pick_count / prob) is slightly off: 
The current correction is
np.log(1/ $q_{i}$ )
where 
$q_{i} = 1 - (1-p_{i})^{S}$ is the sampling probability for in the first attempt.
For n > 1, the prob should be corrected to include the probability of an empty sample 
$P_{0} = \prod_{j} (1 - q_{j})$
as the probability of selection would be $q_i / (1.0 - P_0)$

I think this correction would be generally quite small, but the rows that actually require retries are the rows that would comparatively have a large $P_{0}$.

Second, setting the `prob = 1` for the selected alternatives that pass the fallback method again isn't quite right as the correction factor therefore becomes `np.log(1/prob) = log(1/1) = 0`.  But, the probability of being selected isn't 1 -- it's whatever the selection probability is in the fallback method.

Can we just avoid retrying and if we do not return anything in the first iteration, just select all alternatives that have a non-zero probability? Then their selection prob is 1 and the correction factor becomes moot and we don't waste time and complexity iterating?

I am also wondering whether we should be checking for a "zero probability" error here before we start this looping (or choosing) to catch cases where no alternative is available.

It's difficult for me to say how much any of this matters in practice....

activitysim/core/simulate.py
```diff
@@ -1459,104 +1551,139 @@ def eval_nl(
             column_labels=["alternative", "utility"],
         )
 
-    # exponentiated utilities of leaves and nests
-    nested_exp_utilities = compute_nested_exp_utilities(raw_utilities, nest_spec)
-    chunk_sizer.log_df(trace_label, "nested_exp_utilities", nested_exp_utilities)
+    if state.settings.use_explicit_error_terms:
+        raw_utilities = logit.validate_utils(
+            state, raw_utilities, allow_zero_probs=True, trace_label=trace_label
```
dhensle: should this allow_zero_probs be True? Shouldn't this be passed in from our global/model-level settings?

activitysim/core/random.py
```diff
@@ -245,6 +250,265 @@ def random_for_df(self, df, step_name, n=1):
...
+                positions[row_num] = np.argmax(
+                    utility_row - np.log(-np.log(row_randoms))
+                )
+            else:
+                # Masked positions are set to -inf so they cannot win argmax,
+                # and the gumbel transform is skipped for them entirely.
+                row_mask = masked[row_num]
+                candidate_values = np.full(n_alts, -np.inf, dtype=np.float64)
+                active = ~row_mask
+                if active.any():
```
dhensle: if not active.any() (i.e. there are no available choices), then the argmax below just selects the first one? I am concerned that we aren't correctly catching the "zero probability" case in this function.  We should be reporting zero probabilities and skipping their choices if that setting is enabled.

activitysim/core/random.py
```diff
@@ -245,6 +250,265 @@ def random_for_df(self, df, step_name, n=1):
...
+
+        assert self.step_name
+        assert self.step_name == step_name
+
+        utility_values = utilities.to_numpy()
+        n_rows, n_alts = utility_values.shape
+        positions = np.empty(n_rows, dtype=np.int64)
+
+        if alt_nrs_df is not None:
+            assert alt_nrs_df.shape == utilities.shape
```
dhensle: Suggest we check the actual index (and columns?) here instead of just the shape.

dhensle (changes requested):
Looks good! Documentation is thorough.  Just a couple of things for you to review.

activitysim/abm/models/joint_tour_participation.py
janzill: Done. Replaced all three occurrences and added a comment on usage too.

activitysim/abm/models/trip_scheduling_choice.py
janzill: Ideally yes, but it is not as simple as the idea suggests and therefore not within budget. Something to discuss in the next engineering meeting.

activitysim/core/configuration/base.py
janzill: Currently, destination choice logsums are different for Poisson because of the different meaning of prob (Poisson reports inclusion probabilities). That could be worked around, and in general I don't think the method matters theoretically, only in the sense that some additional things would need to be implemented. I do not see any advantage to this currently, and this is already a very large PR, so I would suggest just requiring MC for estimation. Looking at Jeff's update of the EET branch at the start of this phase, it looks like he added a "estimation needs MC sampling" explicitly (https://github.com/ActivitySim/activitysim/pull/1042/changes#diff-9ec5e393569d30759b9ced82065107dd053d6f1901a4ad20df23825f19fe5f22). @jpn-- anything I missed here?

activitysim/core/interaction_sample.py
janzill: Agreed and done.

activitysim/core/interaction_sample.py
janzill: Regarding if monte_carlo is required, see my answer in base.py. On the validation I agree in principle, but the sample_method can be set globally and per-component in compute_settings, and it looks like enabling estimation comes from a separate config, which is only read on first use in a model.

activitysim/core/interaction_sample_simulate.py
janzill: Done

activitysim/core/simulate.py
janzill: It should not be, good catch. Fixed.

activitysim/core/random.py
janzill: I thought we can just rely on `validate_utils` always being called upstream in `simulate`, `sample_simulate`, etc. On second thought, this is not a good pattern. For MC there is a `utils_to_probs` step outside `make_sample_choices` (the equivalent to `validate_utils`), and then there is an argument called `allow_bad_probs` in `make_choices`. I've added a corresponding argument to `make_choices_explicit_error_term` that mirrors the `allow_bad_probs` functionality. There's also more documentation now to make things clearer.

activitysim/core/random.py
janzill: Done

activitysim/core/interaction_sample.py
janzill: Yes, these are good points and I agree on the math. The variable-number retries also potentially could mess up error term alignment. In practice, these issues did not show up for sample sizes of 30 or above and "peaky-enough" alternative sets, which is what is currently used in all models I know of, but I agree that it's good to fix this. I am a bit hesitant to add all alternatives here because this would slow down things considerably due to mode choice logsums being very slow. However, there is a solution: Choose the sample_size highest-utility alternatives as a fallback set when Poisson sampling came up empty. Those two events are mutually exclusive and the total inclusion probability becomes $prob_i = q_i + P_0 \times \delta_{i \in fallback}$. This solves the correction probability issue that was there before, as well as the EET error term alignment issue. I've added more documentation in both the code and the official documents on this. I've also added a runtime optimization where the second term is only calculated if $P_0$ is larger than $1e-12$ to avoid calculating if an alternative is in the top sample_size values, which requires a sort of the array, see comments in code.

janzill: Regarding checking for zero probabilities, that's handled in `utils to probs`

janzill:
Thanks for the thorough review @dhensle, some good suggestions there. Your sample correction factor for Poisson retries led to another idea which does not require any retries and removes potential error-term misalignment between scenarios, let me know what you think.

activitysim/core/interaction_sample.py
janzill: On another note, Poisson sampling could also be replaced with probabilistic top-k sampling, there is a known closed-form expression for inclusion probabilities and this would be easy to implement and some preliminary tests suggested that it has the same level of stability as Poisson sampling. However, with the above changes I do not think this is necessary.

janzill: fyi @m-richards

activitysim/core/configuration/base.py
jpn--: I think it is reasonable (preferable) to require MC sampling if using estimation mode, for now:

1. The run-to-run stability benefits for EET are entirely moot when running estimation mode.
2. Poisson sampling gets the number of sampled alternatives statistically about right but allows for the number of samples for any single observation to go over the limit, potentially in rare cases by a lot.  ActivitySim handles this OK by managing the data implicitly in a ragged array.  Currently we are using Larch for estimation and Larch allocates memory in rectangular arrays rather than ragged ones, so e.g. if most observations use about 30 alternatives but one oddball gets 60 then the memory usage for estimation doubles.
3. I don't want to add any more new anything to this PR unless absolutely necessary.

We can relax this in the future if desired, in another PR

dhensle (approved):

PR #1077, opened 2026-05-29
Workplace Location Logsum Overwritten in Estimation

### LINKED ISSUE #897: Location choice logsum overwritten with mode choice logsum in estimation mode

dhensle:
We had already fixed #897, but looks like it popped back up again.  I suspect there was a merge error somewhere that caused this to get re-applied.

Tasks:
- [x] Confirm estimation mode outputs workplace location and workplace modechoice logsums
- [x] Add CI test to make sure this doesn't happen again!


jpn--:
I added a test to prevent regressions.  I confirmed that, in the absence of the fix in this PR, the new test fails.  And (obviously) with the fix the test should pass.

PR #1098, opened 2026-07-31
post EET merge clean up

janzill:
Several small fixes and clean ups:

1. Documentation didn't render math equations because mathjax was not enabled
2. After our sampling discussions I realized that there is an edge case when Poisson sampling comes up empty and the  number of available alternatives is smaller than the sample_size. Added a fix and a unit test.
3. Now that the branch is public-facing we should use standard simulation terminology instead of our informal use of monte carlo for sampling from the inverse cdf with a standard uniform, see https://en.wikipedia.org/wiki/Inverse_transform_sampling. This mainly concerns the naming of sampling methods, and I chose "inverse_cdf".

@jpn-- regarding renaming a settings parameter I assumed that it's ok without deprecation  in this case because the parameter was introduced less than 24h ago and there hasn't been a release yet, but let me know if you think otherwise.

### DIFF
--- a/activitysim/abm/models/trip_scheduling_choice.py
+++ b/activitysim/abm/models/trip_scheduling_choice.py
@@ -279,14 +279,28 @@ def run_trip_scheduling_choice(
         ) in chunk.adaptive_chunked_choosers(state, indirect_tours, trace_label):
             # Sort the choosers and get the schedule alternatives
             choosers = choosers.sort_index()
-            # FIXME-EET: For explicit error term choices, we need a stable alternative ID. Currently, we use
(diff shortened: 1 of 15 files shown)


jpn-- (approved):

jpn--:
@janzill I agree no deprecation warnings are needed here.  The two copilot review comments are both minor but I think reasonable.  I tried to simply accept them so as to merge but I think I lack write access to the `outerl` fork.  If you can accept them I will merge.

janzill:
Agreed, they are reasonable so I accepted both, thanks Jeff

PR #1101, opened 2026-08-07
Support reading skims from Parquet files, in addition to OMX

### LINKED ISSUE #970: Support reading skims in Parquet format
**Description**:
Currently, skims are read from OMX files, but based on our experience, storing skims in Parquet format can reduce file size by ~30%. This would significantly reduce storage requirements and improve efficiency.
We already support reading Parquet instead of CSV for the population synthesizer inputs, which has proven very valuable for reducing the size of model inputs. Extending this capability to skims would provide similar benefits.

**Proposed enhancement**:
Add support for reading skims directly from Parquet format.
Maintain backward compatibility with existing OMX inputs.

**Benefits**:
Reduced disk usage.
Faster I/O performance.

jpn--:
Closes #970 

This pull request adds support for reading skim files in Parquet format in addition to the existing OMX format. The changes allow both dense and sparse Parquet files to be auto-detected and mixed with OMX files in the same workflow, with no changes required to existing settings other than using the `.parquet` or `.pq` file extension. The implementation includes updates to documentation, configuration, and the skim loading logic, ensuring compatibility with legacy workflows and Sharrow 2.16+ features.

**Support for Parquet skim files:**

* Added logic to auto-detect and load skim files in Parquet format (with `.parquet` or `.pq` extensions) alongside OMX files. Parquet files can be dense or sparse and are supported by both legacy and Sharrow-based loaders. [[1]](diffhunk://#diff-4436a0a8b98a348f8cd3c009fecdf73291812a015ce4a352748763d9e5ff2ea2R702-R930) [[2]](diffhunk://#diff-4436a0a8b98a348f8cd3c009fecdf73291812a015ce4a352748763d9e5ff2ea2L721-R956) [[3]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fR133-R170)
* Updated documentation and configuration docstrings in `network.py` to describe Parquet skim file support and usage. [[1]](diffhunk://#diff-7cd0464be26f0d6f27b748f986d2dd245a4a4551772497cbeac00b30a07245cbR128-R141) [[2]](diffhunk://#diff-7cd0464be26f0d6f27b748f986d2dd245a4a4551772497cbeac00b30a07245cbL222-R240) [[3]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fL63-R81)

**Skim data loading and handling:**

* Refactored the skim dataset loading logic to merge data from multiple sources, handle time periods, and fill zeros for missing sparse OD pairs to maintain compatibility with legacy behavior.
* Ensured that Parquet-backed datasets are properly handled when storing skims in shared memory, including closing file handles and supporting dask-backed data.

**Codebase and API updates:**

* Added and documented new internal helpers for time period detection, masking, and zero-filling for Parquet files.
* Updated the skim dictionary factory to cache Parquet file metadata and ensure consistent zone ID handling across mixed input files. [[1]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fR102-R105) [[2]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fR133-R170) [[3]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fR206-R213)
* Clarified and expanded docstrings to reflect support for both OMX and Parquet skim files throughout the codebase. [[1]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fL63-R81) [[2]](diffhunk://#diff-d22dff8c0d69e1cdd94ea4b97fe154f1e882ba9cc1a38e2241b78b92faef1a7fL325-R375)


jpn--:
@amsamimi I cannot formally make you a GitHub "reviewer" on this unless you accept the invitation to join the ActivitySim org, but since this addresses the issue you opened, you should definitely review it anyway 😉

amsamimi (approved):
A Sharrow-enabled smoke test completed successfully using the following configuration: 

```
households_sample_size: 200
num_processes: 1
sharrow: true
models:
  - initialize_landuse
  - initialize_households
  - compute_accessibility
```

jpn--:
Thanks for reviewing this @amsamimi.  

At our recent engineering meeting, we also discussed that functional correctness is a necessary but not sufficient criteria for acceptance of new features; performance impacts are also relevant.  While your smoke test is valuable, it would also be important to know if the new code is performing adequately.  For this PR it may be possible to check that simply from the logs of the smoke test -- the main effect is at skim loading, which happens similarly for a full model run or this tiny piece of one.  Can you look and see what the time needed to load parquet skims is, and compare against the equivalent OMX time on your same underlying data?

amsamimi:
@jpn-- @janzill @joecastiglione 

I ran a full-skim test with a sample size of **10,000**, using a **single processor** and **Sharrow**, under two scenarios: **OMX** and **PARQUET**. The memory/runtime profiles are attached, along with the corresponding log files.

**Key Observations:**
| Metric | PARQUET | OMX |
|----------|----------|----------|
| Peak Memory Usage | 63 GB | 22 GB |
| Total Runtime | 23 min | 28 min |
| initialize_landuse | 2.4 min | 7.1 min |

**OMX Version:**
OMX_memory_profile
[OMX_activitysim.log](https://github.com/user-attachments/files/31364645/OMX_activitysim.log)
[OMX_mem.csv](https://github.com/user-attachments/files/31364647/OMX_mem.csv)


**PARQUET Version:**
memory_profile
[PARQUET_activitysim.log](https://github.com/user-attachments/files/31364487/PARQUET_activitysim.log)
[PARQUET_mem.csv](https://github.com/user-attachments/files/31364486/PARQUET_mem.csv)

jpn--:
A few notes:

- The tally of "Writing Skims to Shared Memory" at 21 sec vs 7 min is misleading; parquet is faster but not quite that much faster.  I set about this PR with the singleminded instruction of "faster". This is achieved by reading parquet in parallel streams as fast as possible, but this analysis highlights that we are not piping those data streams directly to the final shared memory destination, we're just loading into whatever memory allocation makes it possible to do with maximum speed.  So while it is technically correct that 21 seconds is the time it takes to get the skim data into shared memory, that is a RAM to RAM copy, and does not include read from disk or decompression.  A more apples-to-apples comparison would include all the read from disk and decompress time, which is effectively the whole `initialize_landuse` step.  Under that measure, we are looking at 2.5 minutes vs 7 minutes.  Still a positive bump but not quite as large.  And that 4.5 minutes difference is effectively the *only* difference between these two runs with respect to runtime.

- It seems likely that with some careful planning about memory use, I should be able to find a solution were we can both have and eat the cake: faster loads, but directly into the shared memory space with no giant temporary array allocation.  Doing this the right way may also deliver improved speed to OMX inputs as well. I already started working towards this in [sharrow#87](https://github.com/ActivitySim/sharrow/pull/87) but I am not done testing and evaluating whether that indeed works as intended.

- I recommend we do not hold this PR until my theorized improvements can be proved out; users can decide if they have the RAM to pay the price to get this acceleration today, and then one day soon via another PR we might make it even better.  I defer to @ActivitySim/product to review the current performance metrics and trade offs outlined above and decide.

amsamimi:
Thanks @jpn-- . I've updated the summary table to avoid any misunderstanding.

Another consideration is that initialize_landuse is not the memory bottleneck. While it has a local peak in memory usage, this does not have any practical consequences, at least for our use case.

PR #1102, opened 2026-08-07
add trip scheduling choice explicit chunking

jpn--:
Replaces and advances #1041.

This pull request adds support for explicit chunk sizing in the `trip_scheduling_choice` model, allowing users to control how data is split during processing. It also introduces comprehensive tests to ensure that explicit chunking works correctly and does not affect model results.

**Explicit chunking support:**

* Added an `explicit_chunk` parameter to `TripSchedulingChoiceSettings`, enabling users to specify a fixed chunk size or a fraction of the dataset for chunking.
* Updated `run_trip_scheduling_choice` to pass the `explicit_chunk` setting to `chunk.adaptive_chunked_choosers`, ensuring the chunking behavior follows user configuration.

**Testing enhancements:**

* Refactored the test for `run_trip_scheduling_choice` to use deep copies and proper random-number management, ensuring reproducibility when chunking is enabled. [[1]](diffhunk://#diff-c5974ec2b4e78c00a672de000b78e3d01c326749b010c7042b2167ab4c6d4eefL231-R239) [[2]](diffhunk://#diff-c5974ec2b4e78c00a672de000b78e3d01c326749b010c7042b2167ab4c6d4eefL249-R259)
* Added a test to verify that explicit chunking divides the data as expected and that results are identical to the default adaptive chunking.

### DIFF
--- a/activitysim/abm/models/trip_scheduling_choice.py
+++ b/activitysim/abm/models/trip_scheduling_choice.py
@@ -276,7 +276,12 @@ def run_trip_scheduling_choice(
             choosers,
             chunk_trace_label,
             chunk_sizer,
-        ) in chunk.adaptive_chunked_choosers(state, indirect_tours, trace_label):
+        ) in chunk.adaptive_chunked_choosers(
+            state,
+            indirect_tours,
+            trace_label,
+            explicit_chunk_size=model_settings.explicit_chunk,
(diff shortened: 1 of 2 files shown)


m-richards:
Thanks @jpn-- for following up on this!

PR #1103, opened 2026-08-09
Faster random numbers

jpn--:
## Summary

This PR adds high-performance, vectorized random-number channels while preserving ActivitySim’s legacy RNG behavior as an option.

Two accelerated modes are available through `rng_channel_type`:

- `pcg64`: PCG64 with robust entropy generation. Better than legacy for large models but still following rigorous "safe" randomness algorithms)
- `sfc64_hash`: SFC64 with lower-overhead hash-based reseeding. Fastest overall for nearly all purposes, but employs short cuts on seeding that are probably fine for large scale simulation, but not rigorously validated as fully uncorrelated random streams to the highest possible levels of confidence
- `legacy`: original `RandomState` implementation and default for backward compatibility

## Key changes

- Adds vectorized uniform, normal, lognormal, choice, Gumbel, and stable-alternative draws.
- Supports reproducible per-row streams, lazy reseeding, step restarts, and selective offset resets.
- Preserves existing output shapes and parameter broadcasting behavior.
- Integrates RNG selection with ActivitySim settings and workflow state.
- Adds cross-channel contract and pipeline regression coverage for all three modes.
- Adds a performance benchmark and manually triggered GitHub Actions workflow.
- Declares `cffi` as a runtime dependency.
- Documents configuration choices, compatibility considerations, and performance tradeoffs.
- Rebases the work onto current `main` and removes unrelated PR scope.

Note: this PR has advanced notably from the last time we looked at it, as the EET branch introduced several new variants of randomness.  I have iterated this on a couple different AI models to get what I believe to be a good result.


jpn--:
This PR was submitted to GPT-6-Astra for re-review, which surfaced 3 additional issues (now solved):

1. **[P1] Restore bounded memory use for EET draws.**  
   The new Gumbel-max implementation materializes the entire chooser × sample × stable-alternative array. The legacy implementation processes one chooser at a time.

   Using the PR’s benchmark, 128 choosers × four samples × 2,048 stable alternatives peaked at **8.13 MiB versus 0.087 MiB** for `simple`. At 1,000 × 30 × 30,000, the shock array alone would require **6.7 GiB**. Stable-uniform and mapped-choice paths have similar allocation growth. Use compiled row-wise or bounded-batch generation, retaining only needed values or winners while preserving stream consumption.

2. **[P2] Handle zero-size draws without errors or stream advancement.**  
   The reshape logic cannot infer a dimension when the requested shape contains zero. Both accelerated modes raise for uniform/Gumbel `n=0`, normal `size=0`, and replacement choice `size=0`; `simple` returns empty arrays.

   Without replacement, zero-size choice returns an empty array **but advances the stream**. Return correctly shaped empty results without consuming draws, and add cross-channel regression tests.

3. **[P2] Validate choice inputs before generating draws.**  
   Size and population handling silently accepts invalid inputs. I reproduced `size=-1` returning four choices per row from a five-item population, `size=1.5` being truncated, and a two-dimensional population being flattened into unexpected results. NumPy rejects these inputs. Validate nonnegative integer dimensions and a one-dimensional population before changing RNG state.

activitysim/core/fast_random/_fast_random.py
```diff
@@ -0,0 +1,2111 @@
...
+        # within _state_bytes.view(np.uint64), in a contiguous block but
+        # not necessarily in the same order.  Find them
+        viewer = self._state_bytes.view(np.uint64)
+        positions = []
+        for j in range(4):
+            target = state_array[j]
+            for k in range(16):
+                if viewer[k] == target:
+                    break
+            if k >= 15:
```
janzill: Should this be `k > 15`?

activitysim/core/fast_random/__init__.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+from ._fast_channel import FastChannel  # noqa: F403
+
+__all__ = (  # noqa: F405
+    # TODO: Add all public symbols here.
```
janzill, line 6: open todo

activitysim/core/configuration/top.py
```diff
@@ -748,6 +748,33 @@ def _check_store_skims_in_shm(self):
     rng_base_seed: Union[int, None] = 0
     """Base seed for pseudo-random number generator."""
 
+    rng_channel_type: Literal["simple", "fast", "faster"] = "simple"
```
janzill, line 751: I find `simple`, `fast`, `faster` misleading. How about `legacy`, `default`, `experimental`?

activitysim/core/fast_random/_fast_random.py
```diff
@@ -0,0 +1,2111 @@
...
+class FastGenerator:
+    def __init__(
+        self, seed: int = 42, bit_gen: Literal["PCG64", "SFC64"] = "PCG64"
+    ) -> None:
+        self._bit_gen_class = bit_gen
+        self._bit_generator = self._new_bit_generator(seed)
+        self._next_uint64 = self._bit_generator.cffi.next_uint64
+        self._next_double = self._bit_generator.cffi.next_double
+        self._state_address = self._bit_generator.cffi.state_address
+        self._state_ptr = _FFI.cast("uint8_t(*)[128]", self._state_address)
```
janzill, line 1884: AI flagged this as a potential problem because it reads 128 bytes from `state_address`, which depends on numpy internal memory layout and currently reads past the object's memory. If numpy's internal layout would change between numpy versions and the scan in the loop below fails, then ActivitySim will fail on import because `FastChannel` is imported in `random.py`, and `entropy` constructs two `FastRandom` at import time. Can we avoid this and only do it when the new RNG is used? That way any potential problems here do not show up when running with the legacy RNG.

activitysim/core/fast_random/_fast_random.py
```diff
@@ -0,0 +1,2111 @@
...
+            if k >= 15:
+                raise ValueError("the state array is not found in raw memory")
+            positions.append(k)
+
+        # Determine where the block of state values is, and verify that it is contiguous
+        self._slice_start = min(positions)
+        self._slice_end = max(positions) + 1
+        if self._slice_start + 4 != self._slice_end:
+            raise ValueError("the state array is not contiguous")
+
```
janzill: AI suggested changes to go with the out-of-range reading above:
```python
        # The four state words occupy a single contiguous 4-word block in the
        # raw buffer, but not necessarily in the same order.  Find the block by
        # matching it as a multiset, then record the permutation within it.
        viewer = self._state_bytes.view(np.uint64)
        wanted = np.sort(state_array)
        for start in range(viewer.size - 3):
            if np.array_equal(np.sort(viewer[start : start + 4]), wanted):
                break
        else:
            raise ValueError(
                "the bit generator state block was not found in raw memory"
            )

        self._slice_start = start
        self._slice_end = start + 4
        block = viewer[start : start + 4]
        self._slice_positions = np.array(
            [int(np.flatnonzero(block == word)[0]) for word in state_array]
        )
```

activitysim/core/fast_random/_fast_random.py
janzill, line 1908:
```suggestion
            else:
```

janzill (approved):
This took a while to review, partly because I wanted to find a way to avoid the large amount of added code and complexity but I don't think there is a way around it within the current framework. In the end, I think we should merge it, I just flagged a couple of minor things.

PR #1110, opened 2026-09-08
Reduce peak memory use in destination choice and large trip models

jpn--:
# Summary
This PR reduces peak and retained memory in large ActivitySim runs, particularly during location choice, tour/trip destination choice, trip scheduling, and trip mode choice.

The main change is to move chunk boundaries outward so that entire high-memory pipelines—not only utility evaluation—operate on bounded chooser subsets. Large sampled-alternative joins, preprocessors, TAZ-to-MAZ expansion, logsum calculations, and final simulation are completed and released one chunk at a time.

It also adds a cross-platform, best-effort mechanism for returning unused native allocator pages to the operating system between large model segments.

# Motivation
Existing evaluator chunking limits the size of individual utility calculations, but several model components still constructed complete merged chooser or sampled-alternative tables before entering those evaluators. At production scale, these intermediate tables can contain millions of rows and many derived columns per worker.
This caused two problems:
- Peak memory could exceed the available RAM even when explicit chunking was enabled.
- Long-lived multiprocess workers retained allocator pages after temporary NumPy and pandas objects were deleted, allowing RSS to remain high across successive model segments.

The in-development Boston full scale model exposed these issues with approximately 3.6 million households, 8.7 million persons, and more than 30 million generated trips.

# Changes
## Bound location and tour-destination logsums
Workplace, school, and tour-destination logsum calculations now:
- Chunk choosers and sampled alternatives before joining person attributes.
- Run the logsum preprocessor and evaluator within each outer chunk.
- Concatenate results in the original sampled-alternative order.
- Disable inner evaluator chunking because the outer loop owns the chunk lifecycle.
This prevents a complete sampled-alternative/person merge and its derived columns from remaining live throughout logsum evaluation.

The logsum utility accepts an optional explicit chunk-size override so callers can distinguish configured outer chunking from disabled inner chunking.

## Bound the complete trip-destination pipeline
When `explicit_chunk` is configured, trip destination now streams each chooser chunk through:
1. Destination sampling.
2. TAZ-to-MAZ expansion.
3. Origin-to-destination and destination-to-primary-destination logsums.
4. Alternative preprocessing.
5. Final destination simulation.

Only the final one-row-per-trip choices are retained between chunks.

Additional changes include:
- Bounding TAZ presampling and MAZ selection together, since MAZ expansion was a significant source of peak memory.
- Moving sampled-alternative/trip merges inside logsum chunks.
- Running alternative preprocessing inside the same chunk as final simulation.
- Preventing nested adaptive chunking by passing zero chunk sizes to inner evaluators.
- Releasing temporary arrays after each chunk and purpose segment.
Estimation and destination-sample-table workflows retain their original whole-segment lifecycle because they intentionally consume the complete sampled table.

If explicit chunking is not configured, trip-destination sampling follows the legacy unchunked path. This preserves compatibility with existing models and lightweight test settings objects.

## Reduce full trip-table copies
Trip scheduling now reuses its existing working trip frame instead of creating a second complete copy near the end of the component. Temporary scheduling columns are removed, and any pre-existing columns used as scratch space are restored.
For school-escort runs, the first working frame is released before reconstructing the complete trip table.

Trip mode choice now:
- Groups the original trips table by primary purpose.
- Merges each purpose segment with only the required tour columns.
- Releases each merged segment after its choices have been collected.
- Avoids retaining a complete trips/tours merged table.

Post-choice annotations temporarily receive `trip_period` for the complete trips table so three-dimensional skim expressions remain correctly aligned. The temporary column is removed afterward, preserving the original table schema.

## Return unused allocator memory
A new `mem.release_memory()` helper:
- Runs Python garbage collection.
- Calls `malloc_trim` on Linux when available.
- Calls `malloc_zone_pressure_relief` on macOS when available.
- Requests working-set trimming on Windows.
- Degrades safely to garbage collection if the platform-specific operation is unavailable.

These calls are advisory and do not affect model behavior if the allocator declines to release memory.

## Supporting production-run fixes
Two supporting fixes were needed for programmatic and containerized benchmarking:
- Programmatic multiprocess runs now forward the tracing run ID in the same way as the CLI.
- Example resources used by `activitysim create` are explicitly included in built packages, avoiding incomplete wheels in Docker or source-only build contexts.

These changes are logically separate from model chunking but support running and validating large production models outside the CLI.

# Compatibility and model results
The changes do not require updates to existing model specifications or prototype regression targets.
- Legacy behavior is preserved when explicit chunking is not configured.
- Estimation and sample-table generation retain their original processing paths.
- Outer chunkers preserve the configured chunk size; inner chunking is disabled to prevent illegal nested adaptive chunking.
- Chunk results are concatenated and checked against the original index order.

Both Sharrow and non-Sharrow execution paths remain supported.

Note: This PR is a result of work for CTPS (Boston MPO) on model optimization, not a consortium funded task.


jpn--:
## abench results

Auto-generated by **abench 0.1.2**.

PR: ActivitySim/activitysim#1110 · Baseline: main

| Run | Result | Elapsed (s) | Change | Peak (GiB) | Change |
|---|---|---:|---:|---:|---:|
| main | valid | 699.33 | +0.0% | 25.728 | +0.0% |
| pr | valid | 501.67 | -28.3% | 25.802 | +0.3% |

Negative changes mean lower elapsed time or memory. Changes are omitted for invalid runs or differing comparison settings. Single runs do not establish statistical significance.

### Configuration and source commits

**main** — MTC main — explicit chunking
Households: 500000; multiprocess: True; processes: 4; Sharrow: True; profile: MTC prototype; platform: None.
- activitysim: ActivitySim/activitysim at `8ca6741c83e44e70fbfe718cb91066dece989805`
- sharrow: ActivitySim/sharrow at `fc175b27d8e0c5d202721c67d96b050e6117b235`

**pr** — MTC PR 1110 — explicit chunking
Households: 500000; multiprocess: True; processes: 4; Sharrow: True; profile: MTC prototype; platform: None.
- activitysim: ActivitySim/activitysim at `51e298a84276813946e1d623c9a5785e078e022f`
- sharrow: ActivitySim/sharrow at `fc175b27d8e0c5d202721c67d96b050e6117b235`

### Memory traces

Memory traces

Blue: total cgroup memory; green dashed: anonymous + shared memory when available. Panels share axes; invalid runs are labeled.

### Component runtimes

Component runtimes

Valid runs only. Mean ± population standard deviation across worker executions, not confidence intervals. Parallel component times must not be summed as wall time.

The interactive HTML report and JSON remain in the experiment output directory.

